PR #2, opened 2015-08-13
fatal error wasn't being triggered until after list(SORT ...) error would happen

codebot:


jacquelinekay:
+1

dirk-thomas:
`get_available_rmw_implementations` should be able to return an empty list if there is no rmw implementation instead of raising a fatal error.

dirk-thomas:
I don't know why you would need a conditional - the following works for me with CMake:

    set(foo "")
    list(SORT foo)

dirk-thomas:
While it is being set in https://github.com/ament/ament_cmake/blob/9048899acba0a095dfbddf7f2645d3c8f2937bfa/ament_cmake_core/cmake/index/ament_index_get_resources.cmake#L60 it looks like that the variable is not defined in the parent scope afterwards.

A simple test proves that:

    set(foo "")
    if(DEFINED foo)
      message("foo defined")
    endif()
    list(SORT foo)
    message("sort ok")

    function(func)
      set(foo "" PARENT_SCOPE)
    endfunction()
    func()
    if(NOT DEFINED foo)
      message("foo NOT defined")
    endif()
    list(SORT foo)

dirk-thomas:
The easiest fix is probable to add the following line after the `ament_index_get_resources()` invocation:

    # ensure that the variable is defined if the function tries to set an empty variable in the parent scope
    set(middleware_implementations "${middleware_implementations}")

dirk-thomas:
Yes, https://github.com/ros2/rmw_implementation/pull/2/files#diff-9e242b1962e7673fca91ac5667e8e57cL27 will result in a fatal error message when `middleware_implementations` is either empty or undefined.

The error message is intended if you ask for the default implementation and there is none. But there should be no fatal error when only asking for a list of available rmw implementations.

PR #3, opened 2015-08-19
add add_executable_for_each_rmw_implementations()

dirk-thomas:
http://ci.ros2.org/job/ros2_batch_ci_linux/226/
http://ci.ros2.org/job/ros2_batch_ci_osx/147/
http://ci.ros2.org/job/ros2_batch_ci_windows/227/

### DIFF
--- a/rmw_implementation/cmake/add_executable_for_each_rmw_implementation.cmake
+++ b/rmw_implementation/cmake/add_executable_for_each_rmw_implementation.cmake
@@ -1,0 +1,78 @@
+# Copyright 2015 Open Source Robotics Foundation, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


esteve:
+1

tfoote:
+1

Is there a recommended way to use this with a gtest target? Running unit tests against all platforms should be our recommendation, it would be good to make that easy.

dirk-thomas:
The function only creates executables (hence `add_executable_*`). A similar function for tests would be possible but also imply a run dependency on e.g. gtest which I would avoid for something like rmw_implementation.

tfoote:
Yeah, this would be an additional helper function. Can we provide an iterator/helper method instead to feed into a foreach block?

dirk-thomas:
I think it would require quite some changes to make the function produce gtest executables and also register them for tests. It is already possible to iterate all generated targets after calling this function. May be that is a better approach to extend its behavior.

tfoote:
Yeah, 

Maybe just a function to load the various middlewares turning this into a one liner: 

```
  # get the rmw implementations ahead of time
  get_available_rmw_implementations(middleware_implementations)
  foreach(middleware_impl ${middleware_implementations})
    find_package("${middleware_impl}" REQUIRED)
  endforeach()
```

Such that we can recommend doing the below and using the variable: 
```
  # create executables
  foreach(middleware_impl ${middleware_implementations})
    ...
```

PR #5, opened 2015-09-08
find_package() in root scope

dirk-thomas:
Invoke `find_package()` in the root scope (rather then in a function scope) to prevent finding the same package repeatedly. E.g. http://ci.ros2.org/job/ros2_batch_ci_linux/320/consoleFull#console-section-74

http://ci.ros2.org/job/ros2_batch_ci_linux/331/consoleFull#console-section-73

PR #6, opened 2015-09-09
fix missed variable name in refactoring

wjwwood:
This was introduced by https://github.com/ros2/rmw_implementation/pull/5.

The result was that no implementation specific binaries were created on my machine.


dirk-thomas:
Sorry, good catch. I think I will update the patch to not rely on the outside variable name but actually pass it as an argument.

wjwwood:
Np, change it how ever you think is best.

dirk-thomas:
I have updated and squashed the commit. I built it locally and it created N executables again.

dirk-thomas:
+1

PR #9, opened 2015-12-08
refactor typesupport

dirk-thomas:
This is work-in-progress.

It currently builds with a single RMW impl. on Linux and OS X.

### DIFF
--- a/rmw_implementation/CMakeLists.txt
+++ b/rmw_implementation/CMakeLists.txt
@@ -1,14 +1,18 @@
 cmake_minimum_required(VERSION 2.8.3)
 
-project(rmw_implementation NONE)
+# must be not NONE since some rmw implementation require compiler checks
+project(rmw_implementation)
 
 find_package(ament_cmake REQUIRED)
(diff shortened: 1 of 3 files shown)


dirk-thomas:
Finally ready for review: *Type Support reloaded*

* http://ci.ros2.org/job/ci_linux/712/
* http://ci.ros2.org/job/ci_osx/590/
* http://ci.ros2.org/job/ci_windows_opensplice/779/

Even a multi-rmw Windows builds works again now: http://ci.ros2.org/job/test_ci_windows_all/ we could merge the RMW specific jobs back into a single one to reduce the overhead again?

While it would be good to first see that building ros2/rosidl#55 on-top of this will work out we should probably move forward anyway, merge this and then work on it in parallel to the rcl work.

wjwwood:
This and the related pr's lgtm

tfoote:
I tested it and it gets me compiling and linking geometry_experimental with the patch in https://github.com/ros2/rclcpp/pull/176 +1 with that patch.

Issue #10, opened 2016-04-14
Trouble specifying default rmw implementation


dhood:
Looking at https://github.com/ros2/rmw/blob/876645761751fc0be3fdaf4bf23d87b408f3e8b4/rmw_implementation_cmake/cmake/get_default_rmw_implementation.cmake, I thought you could specify the default rmw implementation through `ament build --cmake-args -DRMW_IMPLEMENTATION` or the environment variable `RMW_IMPLEMENTATION`

I can't do either, and I think it's because of this: https://github.com/ros2/rmw_implementation/commit/5eff3ff634980e65504dbeaf67a3a00a0e931e3f#diff-e874e8ccc0d49455417259661b3f88f5R17

The configured `rmw_implementation-extras.cmake` is always setting the default rmw implementation to whatever the default was when it was configured.

I want to build the `rclcpp_examples` package with a specific rmw implementation as default, which is different to the default when I built `rmw_implementation`. If I change the part of https://github.com/ros2/rmw_implementation/commit/5eff3ff634980e65504dbeaf67a3a00a0e931e3f which affects `rmw_implementation-extras.cmake` back to how it was, I can set the default rmw implementation through either the environment variable or the cmake args as expected.

Should I be allowed to set the default rmw implementation for `rclcpp_examples` independently to the default used to build `rmw_implementation`? Since `get_default_rmw_implementation` is being used everywhere I thought the answer was yes, but maybe this is the expected behaviour. If so I'd like to add a message that explains why the requested defaults are being ignored.

wjwwood:
You need to rebuild with the `--force-cmake` option. You can choose to only rebuilt the `rmw_implementation` package with `--force-cmake` and that should work as well, at least I think it will.

The location you linked to is expanded each time cmake is run over that package. Since the CMake file for that package did not change, then cmake is not automatically re-run, and the file doesn't get updated.

Other tools like `catkin_make*` and `catkin tools` will watch your command line arguments, and force cmake when you change the cmake variables you are passing. It sounds like that's either not implemented in `ament_tools` or broken, not sure which.

dhood:
did I understand you correctly that you would expect me to be able to run:
```
~/src/ros2examples$ ament build rclcpp_examples --force-cmake-configure --cmake-args -DRMW_IMPLEMENTATION=rmw_fastrtps_cpp
```
(with `export RMW_IMPLEMENTATION=rmw_fastrtps_cpp` for good measure) and see something other than 
```
[...]

PR #11, opened 2016-04-15
Detect if conflicting default rmw implementations have been specified

dhood:
raised from discussion in https://github.com/ros2/rmw_implementation/issues/10

### DIFF
--- a/rmw_implementation/package.xml
+++ b/rmw_implementation/package.xml
@@ -8,8 +8,7 @@
 
   <buildtool_depend>ament_cmake</buildtool_depend>
 
-  <build_depend>rmw_implementation_cmake</build_depend>
-
+  <depend>rmw_implementation_cmake</depend>
   <depend>rmw_connext_cpp</depend>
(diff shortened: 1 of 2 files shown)


wjwwood:
lgtm, I think that will prevent most people from having the wrong expectation when using the `RMW_IMPLEMENATION` CMake/env vars to pick the default.

wjwwood:
@dhood can you run CI over this, at least on Linux just to make sure we don't break master?

dhood:
ok I will

dhood:
I think it's good now http://ci.ros2.org/job/ci_linux/1197/testReport/ (the rclpy tests started failing in build 1191 and I started testing in build 1192)

wjwwood:
@jacquelinekay you might want to have a look at this: http://ci.ros2.org/job/ci_linux/1197/testReport/nose.failure/Failure/runTest/

wjwwood:
lgtm, I tested it locally and it works as expected.

PR #21, opened 2017-04-18
Add security_options to rmw_create_node argument list

mikaelarguedas:
connects to ros2/ros2#339


dirk-thomas:
With the number of arguments growing I suggest to introduce an *options* struct.

PR #22, opened 2017-04-20
rename {c_}utilities to rcutils

wjwwood:
Connects to ros2/c_utilities#17

### DIFF
--- a/rmw_implementation/src/functions.cpp
+++ b/rmw_implementation/src/functions.cpp
@@ -22,7 +22,7 @@
 #include <fstream>
 #include <sstream>
 
-#include "c_utilities/types.h"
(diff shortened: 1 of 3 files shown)


rmw_implementation/src/functions.cpp
```diff
@@ -22,7 +22,7 @@
 #include <fstream>
 #include <sstream>
 
-#include "c_utilities/types.h"
+#include "rcutils/types.h"
```
Karsten1987: including "rcutils/types/string_array.h" , so it stays consistent with the other DDS impl.

wjwwood: c205715

PR #28, opened 2017-10-09
Symplify rmw interface implementation when multi DDS found

jwang11:
The commit use classic c preprocessor approch to handle variadic parameter. By define
RMW interface implementation and handle ARG values inside RMW_INTERFACE_FN macro,
it obviously reduce code lines and complexity of rmw interface implementation.

### DIFF
--- a/rmw_implementation/src/functions.cpp
+++ b/rmw_implementation/src/functions.cpp
@@ -166,23 +166,24 @@ extern "C"
 {
 #endif
 
+#define EXPAND(x) x
 #define ARG_TYPES(...) __VA_ARGS__
 
 #define ARG_VALS0(...)
 #define ARG_VALS1(t1) v1
-#define ARG_VALS2(t2, ...) v2, ARG_VALS1(__VA_ARGS__)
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Thank you for the patch. That indeed significantly decreased the code size.

I applied some minor changes on top in c68550811d43de81b09dd40825052d0f396841ba. I created #29 and triggered some CI builds. It looks like the code doesn't compile on Windows as-is. Could you have a look on how it can be made to work on Windows since we need to support all platforms?

jwang11:
The reason for Windows compile failure lie in the MACRO parse order in a MACRO function and MACRO paramater, like MACRO_FUNC(MACRO_PARAM). In VS, it parse MACRO_FUNC first, but GCC parse MACRO_PARAM firstly. It can be easily resolved #define EXPAND(x) x, which cater both VS and GCC.

dirk-thomas:
Can you commit a patch to this branch? I can then cherry-pick it into the other PR.

mikaelarguedas:
closing this since #29  has been merged.
Thanks @jwang11 for the contribution and for iterating on this!

PR #29, opened 2017-10-12
simplify code relaying all symbols

dirk-thomas:
This slightly updates #28 to:

* update the macro names (underscore between name and number)
* remove semicolons after macro calls
* declare `rmw_init` explicitly to avoid the `init_fn` argument as well as the `FOO_INIT` macro

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3338/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/598/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2668/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3387/)

### DIFF
--- a/rmw_implementation/src/functions.cpp
+++ b/rmw_implementation/src/functions.cpp
@@ -197,12 +197,14 @@ extern "C"
   FunctionSignature func = reinterpret_cast<FunctionSignature>(symbol_ ## symbol_name); \
   return func(arg_values);
 
+#define EXPAND(x) x
+
 #define RMW_INTERFACE_FN(name, ReturnType, error_value, _NR, ...) \
   void * symbol_ ## name = nullptr; \
-  ReturnType name(ARGS_ ## _NR(__VA_ARGS__)) \
+  ReturnType name(EXPAND(ARGS_ ## _NR(__VA_ARGS__))) \
   { \
     CALL_SYMBOL(name, ReturnType, error_value, ARG_TYPES(__VA_ARGS__), \
-      ARG_VALUES_ ## _NR(__VA_ARGS__)); \
+      EXPAND(ARG_VALUES_ ## _NR(__VA_ARGS__))); \
   }
 
 RMW_INTERFACE_FN(rmw_get_implementation_identifier,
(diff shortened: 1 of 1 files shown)


jwang11:
Looks windows build still fail, I'll check, BTW, are you using VS 2017 community ?

dirk-thomas:
> BTW, are you using VS 2017 community ?

We are still using Visual Studio 2015.

jwang11:
find the reason, EXPAND should also apply to ARGS_ and ARG_VALUES, patch soon

dirk-thomas:
@jwang11 Thank you. The new patch looks good. The ongoing Windows build looks good. I will merge this once it is green and the PR got approved.

dirk-thomas:
For now I have just suppressed the warning: https://github.com/ros2/rmw_implementation/pull/29/commits/3d896709ae8210f2d39c83ea7e35df4bf9158b2b#diff-0e10897bf5aa86e527bbfbb0e10759e9R202

dirk-thomas:
Waiting for review.

rmw_implementation/src/functions.cpp
```diff
@@ -182,491 +199,210 @@ extern "C"
...
+  GET_SYMBOL(rmw_get_gid_for_publisher)
+  GET_SYMBOL(rmw_compare_gids_equal)
+  GET_SYMBOL(rmw_service_server_is_available)
 }
 
 void * symbol_rmw_init = nullptr;
 
-rmw_ret_t
-rmw_init(void)
+rmw_ret_t rmw_init()
```
wjwwood: nitpick: why this change?

dirk-thomas: In the first round all functions were refactored into macros and in a second step `rmw_init` was extracted again. I committed 7d4b932976c0e1b34e66da1b5e875db45a071ab1 to avoid the diff.

dirk-thomas:
> I actually find this a bit more difficult to understand. I guess it makes it easier to change the calling mechanism, but I don't expect that to happen very often. Definitely requires more study before you can see what's going on though.

In my opinion the major advantage is the reduction by 260 lines of code as well as not duplicating the function signature of every rmw function anymore.

PR #30, opened 2017-10-20
API to enable log severity setting. 

sriramster:
https://github.com/ros2/rmw/pull/124/ url on rmw.

The patch enables setting log severity from userland applications. The parameter
passed is defined inside rmw/include/types.h

Signed-off-by: Sriram Raghunathan <sriram.max@gmail.com>

Connect to ros2/rmw#124

### DIFF
--- a/rmw_implementation/src/functions.cpp
+++ b/rmw_implementation/src/functions.cpp
@@ -355,6 +355,10 @@ RMW_INTERFACE_FN(rmw_service_server_is_available,
   rmw_ret_t, RMW_RET_ERROR,
   3, ARG_TYPES(const rmw_node_t *, const rmw_client_t *, bool *))
 
+RMW_INTERFACE_FN(rmw_set_log_severity,
+  rmw_ret_t, RMW_RET_ERROR,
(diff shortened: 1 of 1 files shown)


rmw_implementation/src/functions.cpp
```diff
@@ -355,6 +355,10 @@ RMW_INTERFACE_FN(rmw_service_server_is_available,
   rmw_ret_t, RMW_RET_ERROR,
   3, ARG_TYPES(const rmw_node_t *, const rmw_client_t *, bool *))
 
+RMW_INTERFACE_FN(rmw_set_log_severity,
+  rmw_ret_t, RMW_RET_ERROR,
+  1, ARG_TYPES(const rmw_log_severity_t *))
```
dirk-thomas: Since `rmw_log_severity_t` is a primitive type this doesn't need to be passed as `const *`. A plain `rmw_log_severity_t` is sufficient.

sriramster:
@dirk-thomas fixed the comment.

sriramster:
@dirk-thomas Thanks.

PR #31, opened 2017-10-23
*_raw function

Karsten1987:
Connects to ros2/demos#185


rmw_implementation/src/functions.cpp
```diff
@@ -368,6 +380,7 @@ void prefetch_symbols(void)
   GET_SYMBOL(rmw_create_publisher)
   GET_SYMBOL(rmw_destroy_publisher)
   GET_SYMBOL(rmw_publish)
+  GET_SYMBOL(rmw_publish_raw)
```
dirk-thomas: Why is only one of the three new symbols being added here?

PR #32, opened 2017-10-31
use format 3

dirk-thomas:
Connect to ament/ament_package#63

### DIFF
--- a/rmw_implementation/package.xml
+++ b/rmw_implementation/package.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0"?>
-<?xml-model href="http://download.ros.org/schema/package_format2.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
-<package format="2">
+<?xml-model href="http://download.ros.org/schema/package_format3.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="3">
   <name>rmw_implementation</name>
   <version>0.0.3</version>
   <description>The decision which ROS middleware implementation should be used for C++.</description>
@@ -15,13 +15,12 @@
   <depend>libpoco-dev</depend>
   <depend>poco_vendor</depend>
   <depend>rmw_implementation_cmake</depend>
-  <depend>rmw_connext_cpp</depend>
-  <depend>rmw_connext_dynamic_cpp</depend>
   <depend>rmw_fastrtps_cpp</depend>
-  <depend>rmw_opensplice_cpp</depend>
 
   <test_depend>ament_lint_auto</test_depend>
   <test_depend>ament_lint_common</test_depend>
+
+  <group_depend>rmw_implementations</group_depend>
 
   <export>
     <build_type>ament_cmake</build_type>


rmw_implementation/package.xml
```diff
@@ -15,14 +15,13 @@
   <depend>libpoco-dev</depend>
   <depend>poco_vendor</depend>
   <depend>rmw_implementation_cmake</depend>
-  <depend>rmw_connext_cpp</depend>
-  <depend>rmw_connext_dynamic_cpp</depend>
   <depend>rmw_fastrtps_cpp</depend>
-  <depend>rmw_opensplice_cpp</depend>
```
wjwwood: Implicitly the decision here is to release with an explicit dependency on fastrtps only. I guess if we build binaries with support for more we'd add the keys back here?

Perhaps a comment on the fastrtps dependency, indicating that it is there for packaging purposes and is redundant with the group dependency when building from source, would help people interpreting the `package.xml` in the future.

rmw_implementation/package.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0"?>
-<?xml-model href="http://download.ros.org/schema/package_format2.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
-<package format="2">
+<?xml-model href="http://download.ros.org/schema/package_format3.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
```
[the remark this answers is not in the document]
mikaelarguedas: I think right now nothing in the ROS eco-system (1 or 2) requires it. Actually tinyxml(2) just ignores any `<?...?>` tag.
More advanced parsers can use it and complain if the following xml string doesn't validate the provided schema.

rmw_implementation/package.xml
mikaelarguedas: :+1: the absence of comment here is what led to my question of "if a user want to modify the defaut rmw_imp and build only for that one, do they need to modify this manifest?"

rmw_implementation/package.xml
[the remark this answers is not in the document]
dirk-thomas: If you have your editor configured to lint xml files (e.g. in Atom using `linter-xmllint`) it will use the referenced schema. Since the schema is not yet final it isn't available through that URL but you can reference the file with a local `file://` to point to the current version in the branch on the REP repo.

rmw_implementation/package.xml
dirk-thomas: > Implicitly the decision here is to release with an explicit dependency on fastrtps only. I guess if we build binaries with support for more we'd add the keys back here?

Yes, this reflects what we did release in the past betas.

> Perhaps a comment on the fastrtps dependency, indicating that it is there for packaging purposes and is redundant with the group dependency when building from source, would help people interpreting the `package.xml` in the future.

Can you please either propose a wording or commit the comment directly. Thanks.

> if a user want to modify the defaut rmw_imp and build only for that one, do they need to modify this manifest

Nothing has changed here. The user places whatever rmw impl he wants to build into the workspace and if it is more than one and he doesn't use the default (which is FastRTPS if present otherwise the first in alphabetical order) it can be overridden by specifying the environment variable `RMW_IMPLEMENTATION`.

rmw_implementation/package.xml
```diff
@@ -15,14 +15,13 @@
...
   <depend>rmw_implementation_cmake</depend>
-  <depend>rmw_connext_cpp</depend>
-  <depend>rmw_connext_dynamic_cpp</depend>
   <depend>rmw_fastrtps_cpp</depend>
-  <depend>rmw_opensplice_cpp</depend>
 
   <test_depend>ament_lint_auto</test_depend>
   <test_depend>ament_lint_common</test_depend>
 
+  <group_depend>rmw_implementations</group_depend>
```
dirk-thomas: The group name was changed later in #36.

PR #35, opened 2017-11-30
Adding @BannonKing's proposed code

### LINKED ISSUE #33: error from library loader is not as helpful as it could be
When I fail to have the CoreDX environment established, I get an error about being unable to load a library. Unfortunately, the error does not specify the library that failed to load (and it's typically a library several layers deep that failed to load). Please modify the code to look more like this: 
```cpp
    if (library_path.empty()) {
      RMW_SET_ERROR_MSG(("failed to find shared library of rmw implementation. Searched " + env_var).c_str());
      return nullptr;
    }
    try {
      lib = new Poco::SharedLibrary(library_path);
    } catch (Poco::LibraryLoadException& e) {
      RMW_SET_ERROR_MSG(("failed to load shared library of rmw implementation. Exception: " + e.displayText()).c_str());
      return nullptr;
    } catch (...) {
      RMW_SET_ERROR_MSG(("failed to load shared library of rmw implementation: " + library_path).c_str());
      return nullptr;
    }
```

Also note that the library-load code is duplicated in rosidl_typesupport.

kunaltyagi:
Fixes #33

### DIFF
--- a/rmw_implementation/src/functions.cpp
+++ b/rmw_implementation/src/functions.cpp
@@ -305,19 +305,19 @@ RMW_INTERFACE_FN(rmw_trigger_guard_condition,
   rmw_ret_t, RMW_RET_ERROR,
   1, ARG_TYPES(const rmw_guard_condition_t *))
 
-RMW_INTERFACE_FN(rmw_create_waitset,
-  rmw_waitset_t *, nullptr,
+RMW_INTERFACE_FN(rmw_create_wait_set,
+  rmw_wait_set_t *, nullptr,
   1, ARG_TYPES(size_t))
 
-RMW_INTERFACE_FN(rmw_destroy_waitset,
+RMW_INTERFACE_FN(rmw_destroy_wait_set,
   rmw_ret_t, RMW_RET_ERROR,
-  1, ARG_TYPES(rmw_waitset_t *))
+  1, ARG_TYPES(rmw_wait_set_t *))
 
 RMW_INTERFACE_FN(rmw_wait,
   rmw_ret_t, RMW_RET_ERROR,
   6, ARG_TYPES(
     rmw_subscriptions_t *, rmw_guard_conditions_t *, rmw_services_t *,
-    rmw_clients_t *, rmw_waitset_t *, const rmw_time_t *))
+    rmw_clients_t *, rmw_wait_set_t *, const rmw_time_t *))
 
 RMW_INTERFACE_FN(rmw_get_topic_names_and_types,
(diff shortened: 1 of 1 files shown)


dirk-thomas:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3689/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/856/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3020/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3784/)

dirk-thomas:
@kunaltyagi The current patch fails our linters which are run as part of the test. Please update the code and make sure that all tests pass locally.

kunaltyagi:
I ran `src/ament/ament_tools/scripts/ament.py test --only-packages rmw_implementation` and there were no errors. What's the correct process to run the linters?

Output:
```
100% tests passed, 0 tests failed out of 5

Label Time Summary:
copyright     =   0.27 sec (1 test)
cppcheck      =   0.28 sec (1 test)
cpplint       =   0.29 sec (1 test)
lint_cmake    =   0.20 sec (1 test)
linter        =   1.25 sec (5 tests)
uncrustify    =   0.23 sec (1 test)
```

PS: Could we create git hooks to run the linters pre-commit (or pre-push)? This will save some time in integration

dirk-thomas:
> I ran `src/ament/ament_tools/scripts/ament.py test --only-packages rmw_implementation` and there were no errors. What's the correct process to run the linters?

That is the correct command to run the tests of that package. I ran the same command and got the same failing tests as on the referenced Jenkins builds.

I don't know what causes the different behavior in your case. I can only guess maybe you don't have the latest state of all repositories? Can you please make sure that you are using the latest versions of especially the `ament_lint` and `uncrustify` repositories.

> Could we create git hooks to run the linters pre-commit (or pre-push)? This will save some time in integration

Building and running the tests of a package require all dependencies to be available in the current environment. This is not the case in the git hooks. Also a full rebuild would be way too slow to perform it in a git hook.

kunaltyagi:
Fixes the linter issues. Based it from `release-beta3` instead of `master` because of ament error. Rebasing should work :sweat_smile: 

And somehow this issue is closed...

mikaelarguedas:
Arg yeah it looks like something went wrong and this branch has now 0 commit and no code change. @kunaltyagi Do you still have the code somewhere / and if yes, are you still planning to open a PR with that improvement ?

PR #44, opened 2018-07-16
avoid recursive find

dirk-thomas:
In the case where `selected_rmw_implementation` is `rmw_implementation` it shouldn't call `find_package` for it since the entire logic already happens in the CMake config file of `rmw_implementation` so it is already "inside" a `find_package(rmw_implementation)` call.

This only doesn't end in an infinite recursion due to the custom logic in the generated CMake config files which returns early from repeated finds: https://github.com/ament/ament_cmake/blob/6d171d79584319c216e6880c46372224cd3ae6eb/ament_cmake_core/cmake/core/templates/nameConfig.cmake.in#L3-L7

Connect to ros2/ros2#544.

### DIFF
--- a/rmw_implementation/rmw_implementation-extras.cmake.in
+++ b/rmw_implementation/rmw_implementation-extras.cmake.in
@@ -66,13 +66,15 @@ endif()
 
 if(@RMW_IMPLEMENTATION_SUPPORTS_POCO@)
   set(selected_rmw_implementation "@PROJECT_NAME@")
+  # no need to find_package @PROJECT_NAME@
+  # since this code is already part of a find_package call of that package
 else()
   set(selected_rmw_implementation "${default_rmw_implementation}")
   message(STATUS "Using RMW implementation '${selected_rmw_implementation}' as default")
-endif()
-find_package("${selected_rmw_implementation}" REQUIRED)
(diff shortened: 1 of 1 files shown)


dirk-thomas:
CI with 2 RMWs:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4917/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1744/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4082/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4938/)

CI with 1 RMW:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4918/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1745/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4083/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4939/)

dirk-thomas:
While this commit (https://github.com/ament/ament_cmake/pull/146/commits/8003349d5f9d7e10fce0fc4227b171e2538cdea4) was added after CI was triggered the results still hold. Obviously there were no `FATAL_ERROR`s in the builds so the change is only relevant for overlay builds.

PR #53, opened 2019-02-25
Export threading library via extras and not ament_export_libraries to avoid warnings when cross-compiling

esteve:
This PR is based on the work that @anup-pem did internally at Apex. He found that cross compiling ROS2 caused the warning below, which doesn't show up when compiling natively because `pthread` is found on the host but not on the sysroot.

The fix just removes the `ament_export_libraries(pthread)` call and delegates the logic to the extras file.

/cc @filiperinaldi @lmayencourt 

```
--- stderr: velodyne_driver                                  
CMake Warning at /home/anup.pemmaiah/2394/apex_ws/install/share/rmw_implementation/cmake/ament_cmake_export_libraries-extras.cmake:116 (message):
  Package 'rmw_implementation' exports library 'pthread' which couldn't be
  found
:
:
---
```

### DIFF
--- a/rmw_implementation/CMakeLists.txt
+++ b/rmw_implementation/CMakeLists.txt
@@ -61,9 +61,6 @@ else()
   configure_rmw_library(${PROJECT_NAME})
 
   ament_export_libraries(${PROJECT_NAME})
-  if(NOT WIN32)
-    ament_export_libraries(pthread)
-  endif()
 
   install(
     TARGETS ${PROJECT_NAME}
--- a/rmw_implementation/rmw_implementation-extras.cmake.in
+++ b/rmw_implementation/rmw_implementation-extras.cmake.in
@@ -78,3 +78,6 @@ else()
   list(APPEND rmw_implementation_INCLUDE_DIRS ${${selected_rmw_implementation}_INCLUDE_DIRS})
   list(APPEND rmw_implementation_LIBRARIES ${${selected_rmw_implementation}_LIBRARIES})
 endif()
+
+find_package(Threads REQUIRED)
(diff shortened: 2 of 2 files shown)


rmw_implementation/CMakeLists.txt
```diff
@@ -60,11 +60,6 @@ else()
   target_link_libraries(${PROJECT_NAME})
   configure_rmw_library(${PROJECT_NAME})
 
-  ament_export_libraries(${PROJECT_NAME})
```
dirk-thomas: Why should this export be removed?

esteve: Oops, my bad. Good catch. Fixed in https://github.com/ros2/rmw_implementation/compare/8ba2117c1e22eb4dc5082c3d6f3f0a87dc379520..e092ab2fab2b67a48d0f5dcba0ed39d4b1ba3cab

wjwwood:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6268/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2693/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5150/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6109/)

wjwwood:
The test failures seem to be unrelated. The `test_cli` one's I'm not sure about, but the sros2 one is one that I see on our nightly repeated jobs.

esteve:
@dirk-thomas I've addressed your feedback, are you ok with the changes?

esteve:
Thanks @wjwwood and @dirk-thomas for the review.

dirk-thomas:
Thank you for the patch.

PR #59, opened 2019-03-26
cmake: Add `RMW_IMPLEMENTATION_FORCE_POCO`

### LINKED ISSUE #58: cmake: Still build with shared lib loading support, even if there's only one?
At present, the current `CMakeLists.txt` seems a bit too implicit:
https://github.com/ros2/rmw_implementation/blob/c5f5324/rmw_implementation/CMakeLists.txt#L41-L44

Can this be made into an explicit CMake switch, something like `-DFORWARD_RMW_IMPLEMENTATION={OFF,ON}`?

Rationale:
I'm developing on a machine that only has one RMW implementation present.
In debugging stuff like https://github.com/ros2/rcpputils/issues/3 and https://github.com/ros2/rcutils/issues/143, it I couldn't just rebuild `rmw_implementation` to reproduce what I have installed as a binary.

The only way that I can reproduce the binary is either (a) modify `CMakeLists.txt` or (b) install an (unused) RMW implementation, which is odd... ([example patch line](https://github.com/EricCousineau-TRI/repro/commit/cea11026722b340580257564dc49cb0f59b55178#diff-19a9c16469e9b7649ce20f547b993418R12))

I should be able reproduce this with my installed packages?

EricCousineau-TRI:
Resolves #58

\cc @dirk-thomas

---
This change is [Reviewable](https://reviewable.io/reviews/ros2/rmw_implementation/59)

### DIFF
--- a/rmw_implementation/CMakeLists.txt
+++ b/rmw_implementation/CMakeLists.txt
@@ -35,30 +35,34 @@ foreach(impl IN LISTS RMW_IMPLEMENTATIONS)
 endforeach()
 
 message(STATUS "")
-message(STATUS "Using default rmw implementation: '${RMW_IMPLEMENTATION}'")
+message(STATUS "Default rmw implementation: '${RMW_IMPLEMENTATION}'")
 message(STATUS "")
 
-set(RMW_IMPLEMENTATION_FORCE_POCO OFF CACHE BOOL
-    "Force using poco for loading shared libraries, even if there is only one \
-     implementation available")
+option(RMW_IMPLEMENTATION_FORCE_DYNAMIC_LOADING
+    "Force use of dynamic loading (via poco), even if there is only one implementation available"
+    OFF)
 
-set(RMW_IMPLEMENTATION_SUPPORTS_POCO FALSE)
-# if poco is not forced and only a single rmw impl. is available or poco is not available
-# this package will directly reference the default rmw impl.
-if(RMW_IMPLEMENTATION_FORCE_POCO OR (RMW_IMPLEMENTATIONS MATCHES ";" AND POCO_FOUND))
+# Nominally use dynamic loading if poco is available.
+set(RMW_IMPLEMENTATION_SUPPORTS_POCO Poco_FOUND)
+# Determine if we're going to actually support dynamic loading.
+if(RMW_IMPLEMENTATION_FORCE_DYNAMIC_LOADING)
+  # Force usage.
+  if(NOT Poco_FOUND)
+    message(FATAL_ERROR "Dynamic loading forced, but poco is not found!")
+  endif()
   set(RMW_IMPLEMENTATION_SUPPORTS_POCO TRUE)
+elseif(NOT RMW_IMPLEMENTATIONS MATCHES ";")
+  message(STATUS "Only a single rmw implementation found; disabling dynamic loading")
+  set(RMW_IMPLEMENTATION_SUPPORTS_POCO FALSE)
 endif()
 
 if(NOT RMW_IMPLEMENTATION_SUPPORTS_POCO)
-  message(STATUS "Not using poco for runtime override; directly referencing '${RMW_IMPLEMENTATION}'")
+  message(STATUS "Dynamic loading DISABLED; directly referencing '${RMW_IMPLEMENTATION}'")
   message(STATUS "")
 
 else()
-  message(STATUS "Support runtime override with any of the available rmw implementations")
+  message(STATUS "Dynamic loading enabled with any of the available rmw implementations")
   message(STATUS "")
-  if(NOT Poco_FOUND)
-    message(FATAL_ERROR "Configuration desired poco, but it is not found!")
-  endif()
 
   find_package(rmw REQUIRED)
 
--- a/rmw_implementation/CMakeLists.txt
+++ b/rmw_implementation/CMakeLists.txt
@@ -39,11 +39,11 @@ message(STATUS "Default rmw implementation: '${RMW_IMPLEMENTATION}'")
 message(STATUS "")
 
 option(RMW_IMPLEMENTATION_FORCE_DYNAMIC_LOADING
-    "Force use of dynamic loading (via poco), even if there is only one implementation available"
-    OFF)
+  "Force use of dynamic loading (via poco), even if there is only one implementation available"
+  OFF)
 
 # Nominally use dynamic loading if poco is available.
-set(RMW_IMPLEMENTATION_SUPPORTS_POCO Poco_FOUND)
+set(RMW_IMPLEMENTATION_SUPPORTS_POCO ${Poco_FOUND})
 # Determine if we're going to actually support dynamic loading.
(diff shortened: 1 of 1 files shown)


rmw_implementation/CMakeLists.txt
```diff
@@ -38,17 +38,30 @@ message(STATUS "")
...
-  set(RMW_IMPLEMENTATION_SUPPORTS_POCO FALSE)
+set(RMW_IMPLEMENTATION_FORCE_POCO OFF CACHE BOOL
+    "Force using poco for loading shared libraries, even if there is only one \
+     implementation available")
 
-else()
+set(RMW_IMPLEMENTATION_SUPPORTS_POCO FALSE)
+# if poco is not forced and only a single rmw impl. is available or poco is not available
+# this package will directly reference the default rmw impl.
+if(RMW_IMPLEMENTATION_FORCE_POCO OR (RMW_IMPLEMENTATIONS MATCHES ";" AND POCO_FOUND))
```
dirk-thomas: Use consistent case of CMake variables: `Poco_FOUND`.

wjwwood:
I'm just curious why `_POCO` is in the name? The fact that we use poco to implement the shared library loading (rather than using a different dependency or using `dlopen` directly) is a pretty specific implementation detail. We've already had people asking us to drop the dependency on poco and do this some other way. I wouldn't be surprised if that changed in the near future and in that case the variable name would not make sense any longer. I don't think it should be part of the option name.

Googling briefly, I think the generic term for this is "dynamic loading" (https://en.wikipedia.org/wiki/Dynamic_loading), so maybe `RMW_IMPLEMENTATION_FORCE_DYNAMIC_LOADING` is a better option name.

EricCousineau-TRI:
> [...] so maybe RMW_IMPLEMENTATION_FORCE_DYNAMIC_LOADING is a better option name.

SGTM!

EricCousineau-TRI:
Er, also, how do y'all prefer reviews / commit curation?
Do you prefer force pushes with curated commits, or accumulate revision commits until the end?

rmw_implementation/CMakeLists.txt
EricCousineau-TRI: Done - oops!

rmw_implementation/CMakeLists.txt
```diff
@@ -38,17 +38,30 @@ message(STATUS "")
 message(STATUS "Using default rmw implementation: '${RMW_IMPLEMENTATION}'")
 message(STATUS "")
 
-# if only a single rmw impl. is available or poco is not available
-# this package will directly reference the default rmw impl.
-if(NOT RMW_IMPLEMENTATIONS MATCHES ";" OR NOT Poco_FOUND)
-  set(RMW_IMPLEMENTATION_SUPPORTS_POCO FALSE)
+set(RMW_IMPLEMENTATION_FORCE_POCO OFF CACHE BOOL
```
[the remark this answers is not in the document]
EricCousineau-TRI: Done.

EricCousineau-TRI:
> The logic / control flow is pretty hard to read atm. I would suggest to simplify it as follows:

Done, with minor adjustment for logic of `*FOUND*` + `*FORCE*`.

BTW, it would be really nice to make `FORCE` a tri-state setting: null / unspecified implies use implicit behavior; on implies must be enabled, off means disabled. This avoids the need to try to curate the discovered packages on your system if you have a desired configuration.

Not a huge deal, but may simplify things future usages.
Worth considering or now, should we just punt?

@wjwwood Made the local change for `_FORCE` for "dynamic loading" as a term.

I can rename `RMW_IMPLEMENTATION_SUPPORTS_POCO` as well, if it's scope is local to this package.

EricCousineau-TRI:
> Do you prefer force pushes with curated commits [...]

Sorry! Just found my answer here: https://index.ros.org/doc/ros2/Contributing/Developer-Guide/#pull-requests

rmw_implementation/CMakeLists.txt
```diff
@@ -35,20 +35,37 @@ foreach(impl IN LISTS RMW_IMPLEMENTATIONS)
...
-message(STATUS "Using default rmw implementation: '${RMW_IMPLEMENTATION}'")
+message(STATUS "Default rmw implementation: '${RMW_IMPLEMENTATION}'")
 message(STATUS "")
 
-# if only a single rmw impl. is available or poco is not available
-# this package will directly reference the default rmw impl.
-if(NOT RMW_IMPLEMENTATIONS MATCHES ";" OR NOT Poco_FOUND)
+option(RMW_IMPLEMENTATION_FORCE_DYNAMIC_LOADING
+    "Force use of dynamic loading (via poco), even if there is only one implementation available"
+    OFF)
```
dirk-thomas: This should be 2 space indentation.

rmw_implementation/CMakeLists.txt
```diff
@@ -35,20 +35,37 @@ foreach(impl IN LISTS RMW_IMPLEMENTATIONS)
...
+
+# Nominally use dynamic loading if poco is available.
+set(RMW_IMPLEMENTATION_SUPPORTS_POCO Poco_FOUND)
```
dirk-thomas: The value should be the evaluated variable - not the variable name.

rmw_implementation/CMakeLists.txt
```diff
@@ -35,20 +35,37 @@ foreach(impl IN LISTS RMW_IMPLEMENTATIONS)
...
+# Determine if we're going to actually support dynamic loading.
+if(RMW_IMPLEMENTATION_FORCE_DYNAMIC_LOADING)
+  # Force usage.
+  if(NOT Poco_FOUND)
+    message(FATAL_ERROR "Dynamic loading forced, but poco is not found!")
+  endif()
+  set(RMW_IMPLEMENTATION_SUPPORTS_POCO TRUE)
+elseif(NOT RMW_IMPLEMENTATIONS MATCHES ";")
+  message(STATUS "Only a single rmw implementation found; disabling dynamic loading")
```
dirk-thomas: I would remove this message since it is redundant information with the message in line 60.

rmw_implementation/CMakeLists.txt
```diff
@@ -35,20 +35,37 @@ foreach(impl IN LISTS RMW_IMPLEMENTATIONS)
...
   set(RMW_IMPLEMENTATION_SUPPORTS_POCO FALSE)
+endif()
+
+if(NOT RMW_IMPLEMENTATION_SUPPORTS_POCO)
+  message(STATUS "Dynamic loading DISABLED; directly referencing '${RMW_IMPLEMENTATION}'")
```
dirk-thomas: Please use lower case for `DISABLED` - no need to "yell" the word :wink:.

dirk-thomas:
Multiple RMWs: [Build Status](https://ci.ros2.org/job/ci_linux/6532/)
Single RMW: [Build Status](https://ci.ros2.org/job/ci_linux/6533/)

dirk-thomas:
Thanks for the patch.

PR #60, opened 2019-04-03
QoS Events and Liveliness

nburek:
This change is to allow events that occur in the rmw layer related to missed deadlines and liveliness policies to be surfaced up to a user. It adds a new parameter to some existing functions and also adds several new functions in support of the liveliness QoS setting and taking events in general. 

Connects to ros2/rmw#171
Related to https://github.com/ros2/design/pull/212


nburek:
@wjwwood This is ready for review. Can you please update the label? Thank you.

wjwwood:
Looks like something may have gone wrong here...

Make sure you `git fetch origin` and use `origin/master` rather than just `master` when you rebase.

PR #63, opened 2019-09-04
Add CycloneDDS

rotu:
CycloneDDS already works at runtime when setting `RMW_IMPLEMENTATION=rmw_cyclonedds_cpp`. This PR prevents build errors like the following if you have set `RMW_IMPLEMENTATION=rmw_cyclonedds_cpp` at build time
```
colcon build --packages-select rplidar_ros
Starting >>> rplidar_ros
--- stderr: rplidar_ros                         
CMake Error at /opt/ros/dashing/share/rmw_implementation/cmake/rmw_implementation-extras.cmake:47 (message):
  The RMW implementation has been specified as 'rmw_cyclonedds_cpp' through
  the environment variable 'RMW_IMPLEMENTATION', however it is not in the
  list of supported rmw implementations, which was specified when the
  'rmw_implementation' package was built.
Call Stack (most recent call first):
  /opt/ros/dashing/share/rmw_implementation/cmake/rmw_implementationConfig.cmake:38 (include)
  /opt/ros/dashing/share/rcl/cmake/ament_cmake_export_dependencies-extras.cmake:21 (find_package)
  /opt/ros/dashing/share/rcl/cmake/rclConfig.cmake:38 (include)
  /opt/ros/dashing/share/rclcpp/cmake/ament_cmake_export_dependencies-extras.cmake:21 (find_package)
  /opt/ros/dashing/share/rclcpp/cmake/rclcppConfig.cmake:38 (include)
  /opt/ros/dashing/share/ament_cmake_auto/cmake/ament_auto_find_build_dependencies.cmake:67 (find_package)
  CMakeLists.txt:29 (ament_auto_find_build_dependencies)


---
Failed   <<< rplidar_ros	[ Exited with code 1 ]
```

### DIFF
--- a/rmw_implementation/package.xml
+++ b/rmw_implementation/package.xml
@@ -18,6 +18,7 @@
   <build_depend>rmw_connext_cpp</build_depend>
   <build_depend>rmw_fastrtps_cpp</build_depend>
   <build_depend>rmw_opensplice_cpp</build_depend>
+  <build_depend>rmw_cyclonedds_cpp</build_depend>
   <!-- end of group dependencies added for bloom -->
 
   <depend>libpoco-dev</depend>


dirk-thomas:
Please add a description to your PR. Why do you want to make this change?

CycloneDDS uses the introspection typesupport and doesn't need a custom typesupport. Why can you not simply install the CycloneDDS packages or build them from source without this change?

rotu:
@dirk-thomas updated description.

dirk-thomas:
Atm CycloneDDS is not a supported RMW implementation and not mentioned in the REP. Hence we don't list it as a dependency.

I don't think it is possible to list every existing RMW implementation as an explicit build dependency here. That is what the group dependency is used for (to ensure all possible RMW implementations are built before this package). Maybe we can modify the check instead to take group dependencies into account.

rotu:
AFAIK, there are only 4 RMW implementations that work today: FastRTPS, OpenSplice, Connext, CycloneDDS. And a fourth, CoreDX, which is [currently broken](https://github.com/tocinc/rmw_coredx/issues/32). So it's certainly possible to list them, at least at this point in time.

As for group dependencies, there's a comment in package.xml that seems to explain why that wouldn't work:
https://github.com/ros2/rmw_implementation/blob/cf709e83e5d34291c52145cebf9381c7cb0f2dac/rmw_implementation/package.xml#L14-L17

dirk-thomas:
> AFAIK, there are only 4 RMW implementations that work today: FastRTPS, OpenSplice, Connext, CycloneDDS. And a ~~fourth~~ fifth, CoreDX, which is [currently broken](https://github.com/tocinc/rmw_coredx/issues/32). So it's certainly possible to list them, at least at this point in time.

There are actually more than 10 implementation (and that only counts the ones we know about). Therefore I don't think this is feasible.

> As for group dependencies, there's a comment in package.xml that seems to explain why that wouldn't work:

As the comment says this is about `bloom` which doesn't apply here.

Please try the patch proposed in #65 which considers the list of available RMW impl. at runtime rather than only considering the ones which where available when `rmw_implementation` was built.

rotu:
> There are actually more than 10 implementation (and that only counts the ones we know about).

Where? I'm doing a comparative analysis of RMW implementations, and I'd love to try some of these out!

> As the comment says this is about bloom which doesn't apply here.

I didn't quite understand that comment. Nor why we have a whitelist at the time of building `rmw_implementation`. I assumed that whitelist served a purpose, but given your patch, I guess not!

Trying out your patch right now, as it seems a much better alternative moving forward!

rotu:
Closing in deference to #65

PR #65, opened 2019-09-04
consider available rmw implementations at runtime

dirk-thomas:
Alternative to ros2/rmw_implementation#63.

### DIFF
--- a/rmw_implementation/rmw_implementation-extras.cmake.in
+++ b/rmw_implementation/rmw_implementation-extras.cmake.in
@@ -16,7 +16,7 @@
 
 find_package(rmw_implementation_cmake REQUIRED)
 set(default_rmw_implementation "@RMW_IMPLEMENTATION@")
-set(available_rmw_implementations "@RMW_IMPLEMENTATIONS@")
+get_available_rmw_implementations(available_rmw_implementations)
 
 if(NOT "${RMW_IMPLEMENTATION}" STREQUAL "")
   set(requested_rmw_implementation ${RMW_IMPLEMENTATION})
@@ -24,9 +24,8 @@ if(NOT "${RMW_IMPLEMENTATION}" STREQUAL "")
     if(NOT requested_rmw_implementation IN_LIST available_rmw_implementations)
       message(FATAL_ERROR
         "The RMW implementation has been specified as "
-        "'${requested_rmw_implementation}' through CMake, "
-        "however it is not in the list of supported rmw implementations, "
-        "which was specified when the 'rmw_implementation' package was built.")
+        "'${requested_rmw_implementation}' through CMake, however it is not "
+        "registered in the resource index as a known rmw implementation.")
     endif()
   else()
     if(NOT requested_rmw_implementation STREQUAL default_rmw_implementation)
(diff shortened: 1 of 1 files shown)


rotu:
Any reason to keep the build-time dependency on connext/fastrtps/opensplice?

Also, don't forget to remove this now-obsolete check https://github.com/ros2/rmw_implementation/blob/1ca4808d4c4b5cb272038fe2aee0e29f335cdfee/rmw_implementation/CMakeLists.txt#L54-L56

dirk-thomas:
> Any reason to keep the build-time dependency on connext/fastrtps/opensplice?

The comment about `bloom` still applies to them: https://github.com/ros2/rmw_implementation/pull/63#issuecomment-528105897

> Also, don't forget to remove this now-obsolete check

That check is not obsolete. If there is only a single RMW implementation available at build time of `rmw_implementation` then it shouldn't use poco and instead directly link against that RMW implementation (which makes choosing the implementation using an environment variable impossible).

rotu:
1. I don't understand the comment about bloom, so I'm gonna assume you've thought about this and that you're right.
2. If there is 0 or 1 RMW implementations when the underlay is built that doesn't mean there will be a single implementation when rmw_implementation is actually used.
3. This fix seems to work! I'm closing my PRs.

dirk-thomas:
> It seems like maybe it would make sense to have an option like RMW_IMPLEMENTATION_NO_DYNAMIC_LOADING like the existing RMW_IMPLEMENTATION_FORCE_DYNAMIC_LOADING option.

That would be an option to allow forcing dynamic loading even if only one rmw impl is present. Please consider to create a PR for that enhancement.

dirk-thomas:
> It seems like maybe it would make sense to have an option like RMW_IMPLEMENTATION_NO_DYNAMIC_LOADING like the existing RMW_IMPLEMENTATION_FORCE_DYNAMIC_LOADING option.

That would be an option to force static dynamic loading even if more than one rmw impl is present. On the other hand if you really care about that optimization you would probably not have more than one rmw impl. in your workspace in the first place. Anyway please consider to create a PR for that enhancement if you think it is valuable.

rotu:
> Anyway please consider to create a PR for that enhancement if you think it is valuable.

I guess I could, though I still don’t get why the special case is even needed when there is no more than one package.

It really just strikes me as a mistake that there are now two lists of packages that can affect the build: one calculated when rmw_implementation is built and one calculated when downstream consuming packages are built. I don’t see the necessity of *both* `RMW_IMPLEMENTATIONS` and `available_rmw_implementations`, and I can see it causing subtle build bugs down the line.

dirk-thomas:
CI builds (up to `test_communication`) with 2 rmw impls:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8067/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4075/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6573/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7939/)

CI builds (up to `test_communication`) with 1 rmw impl:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8068/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4076/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6574/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7940/)

dirk-thomas:
> It really just strikes me as a mistake that there are now two lists of packages that can affect the build: one calculated when rmw_implementation is built and one calculated when downstream consuming packages are built. I don’t see the necessity of *both* `RMW_IMPLEMENTATIONS` and `available_rmw_implementations`, and I can see it causing subtle build bugs down the line.

I agree that having two different kind of checks is awkward. I just don't see what the alternative should be if the requirement is that the current uses cases should continue to be supported:

* with a single RMW impl. present at build time `rmw_implementation` should directly link against it (for efficiency and simplicity)
* for your use case you want to be able to use an RMW implementation which wasn't available a build time of `rmw_implementation` and select it using the environment variable

Please feel free to propose an alternative approach.

rotu:
Proposed alternate approach in #67

dirk-thomas:
Closing in favor of the merged PR #67.

PR #68, opened 2019-09-20
[Dashing] Make middleware selection more independent of build-tim…

rotu:
Dashing version of #67

### DIFF
--- a/rmw_implementation/CMakeLists.txt
+++ b/rmw_implementation/CMakeLists.txt
@@ -15,9 +15,6 @@ find_package(ament_cmake REQUIRED)
 
 find_package(rcutils REQUIRED)
 
-# provides FindPoco.cmake and Poco on platforms without it
-find_package(poco_vendor)
-find_package(Poco COMPONENTS Foundation)
 find_package(rmw_implementation_cmake REQUIRED)
 
 if(BUILD_TESTING)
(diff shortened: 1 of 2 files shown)


rotu:
rebased to exclude accidentally included commits from master

dirk-thomas:
Waiting until #67 has been accepted into `master`.

dirk-thomas:
I am not convinced this change should be backported. It breaks previously working CMake options. Maybe someone else from the @ros2/team can comment on this.

rotu:
It’s a PITA to use Dashing binaries with CycloneDDS without this patch. Your call

dirk-thomas:
I do understand the benefits of the patch. That is why it has already landed on the master branch.

Please also understand that Dashing is an already released distro and breaking API / behavior in a already released distro is something we do **not** take lightly. Also at the time of the release of Dashing CycloneDDS was not part of it - so this is also not about fixing a bug.

rotu:
I’m not convinced it’s not a bug, maybe in design maybe in implementation. Again, the PR is done. Not my call to accept or reject it

rotu:
It's easy enough to make the old build flag still work. I don't think it breaks ABI, but I don't understand enough about ABI to know.

wjwwood:
> It's easy enough to make the old build flag still work.

If it can be made such that no existing option/user is impacted, that would be much more palatable. But maybe wait until others chime in before spending the effort to do that. Just knowing it may be possible is good though.

> I don't think it breaks ABI, but I don't understand enough about ABI to know.

ABI is specifically about compatibility in C++'s binary interface between libraries, but in this case I'm (ab)using it as "ABI compatibility means no intervention for existing users in anyway after updating". That means if they're using an option and expect a certain behavior we shouldn't violate that expectation if at all possible.

dirk-thomas:
> Can someone summarize the impact on existing use cases? Like what can you do now that would not work after?

The patch introduces some CMake options and removes other CMake options. If users use the CMake options which are removed their workflow will not only break but also silently since CMake will just ignore the not defined option.

> It's easy enough to make the old build flag still work.

That would resolve the problem of breaking API / behavior.

dirk-thomas:
> It's easy enough to make the old build flag still work.

@rotu Can you update this PR to maintain compatibility with the previous options?

rotu:
@dirk-thomas Done

nuclearsandwich:
> I will leave it up to you to merge it (since I am not sure if you want any extra testing to happen).

Thanks. We just had a sync so if we get this merged and released soon we'll have plenty of time for it to soak on testing. I've added it to the patch release board and put time on my board next week for testing before the release.

nuclearsandwich:
I haven't seen a Dashing CI run in a while. We can definitely expect some linter issues as Dashing doesn't get updates for updates to linters.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8674/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4514/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7083/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8595/)

PR #70, opened 2019-09-25
remove unneeded line from CMakeLists

rotu:
I don't think we need to tweak the link_directories to get this Poco to link properly

### DIFF
--- a/rmw_implementation/CMakeLists.txt
+++ b/rmw_implementation/CMakeLists.txt
@@ -54,7 +54,6 @@ else()
   find_package(Poco REQUIRED COMPONENTS Foundation)
   find_package(rmw REQUIRED)
 
-  link_directories(${Poco_LIBRARY_DIR})
   add_library(${PROJECT_NAME} SHARED
     src/functions.cpp)
   ament_target_dependencies(${PROJECT_NAME}


dirk-thomas:
This builds on top of #67.

As [mentioned in that ticket](https://github.com/ros2/rmw_implementation/pull/67#discussion_r328341384) please consider to provide the additional information so that we can be certain that the `link_directories(${Poco_LIBRARY_DIR})` call can be safely removed.

rotu:
What additional information is needed? Correctly building on target platforms should prove that this PR is correct. Please run CI to confirm.

dirk-thomas:
> What additional information is needed?

It helps to understand why it was used and necessary in the past, what has changed in the meantime, and confirms it can be safely removed now.

> Correctly building on target platforms should prove that this PR is correct. Please run CI to confirm.

Passing builds only show that it works on the exact configurations used but doesn't provide us with any understanding.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8232/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4203/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6712/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8116/)

rotu:
You made that change, so you tell me. It looks to me like it was never necessary.

https://github.com/ros2/rmw_implementation/commit/52a46ba1df6b51b8d0ae95f62926e49d2ece8b1f#diff-0c408aa74d7519e29ea65afd30874fcf

https://github.com/ros2/rmw_implementation/pull/17

> Passing builds only show that it works on the exact configurations used but doesn't provide us with any understanding.

True but (1) that applies to everything (2) CI gives the best assurance that this works across supported build configurations.

dirk-thomas:
> True but (1) that applies to everything (2) CI gives the best assurance that this works across supported build configurations.

I like to understand *why* things work / don't work and what has changed.

I am pretty sure for earlier versions of Poco which didn't bundle a CMake config this variable was provided by either a distro specific patch or a vendor-like wrapper package.

Anyway as of version 1.6.0 Poco include a CMake config file (see https://github.com/pocoproject/poco/blob/poco-1.6.0-release/cmake/PocoConfig.cmake.in). So it seems to be fine to remove.

dirk-thomas:
With #67 merged this needs a rebase.

rotu:
Rebased

PR #72, opened 2019-10-04
Added functions to get qos policies for publishers and subscribers to a topic

jaisontj:
As a part of implementing [this](https://github.com/ros2/rmw/issues/151) feature. The RMW layer needs to expose apis to retrieve a list of publishers and subscribers along with their QoS policies for a given topic.

- rmw_get_qos_for_subscribers to get the qos policies for subscribers to
a topic
- rmw_get_qos_for_publishers to a get the qos policies for publishers to
a topic

Merge this PR after [this](https://github.com/ros2/rmw/pull/186) is approved as the new functions depend on the newly created structure `rmw_participants_t`

### DIFF
--- a/rmw_implementation/src/functions.cpp
+++ b/rmw_implementation/src/functions.cpp
@@ -535,7 +535,7 @@ RMW_INTERFACE_FN(rmw_set_log_severity,
   1, ARG_TYPES(rmw_log_severity_t))
 
 RMW_INTERFACE_FN(rmw_get_publishers_info_by_topic,
-  rmw_ret_t, RMW_RET_UNSUPPORTED,
+  rmw_ret_t, RMW_RET_ERROR,
   5, ARG_TYPES(
     const rmw_node_t *,
     rcutils_allocator_t *,
@@ -544,7 +544,7 @@ RMW_INTERFACE_FN(rmw_get_publishers_info_by_topic,
     rmw_topic_info_array_t *))
 
 RMW_INTERFACE_FN(rmw_get_subscriptions_info_by_topic,
-  rmw_ret_t, RMW_RET_UNSUPPORTED,
+  rmw_ret_t, RMW_RET_ERROR,
   5, ARG_TYPES(
     const rmw_node_t *,
     rcutils_allocator_t *,
@@ -626,9 +626,7 @@ void prefetch_symbols(void)
   GET_SYMBOL(rmw_service_server_is_available)
   GET_SYMBOL(rmw_set_log_severity)
   GET_SYMBOL(rmw_get_publishers_info_by_topic)
-  rmw_reset_error();
   GET_SYMBOL(rmw_get_subscriptions_info_by_topic)
-  rmw_reset_error();
 }
 
 void * symbol_rmw_init = nullptr;
--- a/rmw_implementation/src/functions.cpp
+++ b/rmw_implementation/src/functions.cpp
@@ -33,7 +33,7 @@
 #include "rmw/names_and_types.h"
 #include "rmw/get_node_info_and_types.h"
 #include "rmw/get_service_names_and_types.h"
-#include "rmw/get_topic_info.h"
+#include "rmw/get_topic_endpoint_info.h"
 #include "rmw/get_topic_names_and_types.h"
 #include "rmw/rmw.h"
 
@@ -542,7 +542,7 @@ RMW_INTERFACE_FN(rmw_get_publishers_info_by_topic,
     rcutils_allocator_t *,
     const char *,
     bool,
-    rmw_topic_info_array_t *))
+    rmw_topic_endpoint_info_array_t *))
 
 RMW_INTERFACE_FN(rmw_get_subscriptions_info_by_topic,
   rmw_ret_t, RMW_RET_ERROR,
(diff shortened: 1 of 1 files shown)


jaisontj:
The build will fail until [this](https://github.com/ros2/rmw/pull/186) is approved.

jaisontj:
@ivanpauno Related to [this](https://github.com/ros2/rmw/pull/186)

rmw_implementation/src/functions.cpp
```diff
@@ -534,6 +534,24 @@ RMW_INTERFACE_FN(rmw_set_log_severity,
...
+  rmw_ret_t, RMW_RET_UNSUPPORTED,
+  5, ARG_TYPES(
+    const rmw_node_t *,
+    rcutils_allocator_t *,
+    const char *,
+    bool,
+    rmw_topic_info_array_t *))
+
+RMW_INTERFACE_FN(rmw_get_subscriptions_info_by_topic,
+  rmw_ret_t, RMW_RET_UNSUPPORTED,
```
jaisontj: Returning `RMW_RET_UNSUPPORTED` as I think this is more apt than `RMW_RET_ERROR`

ivanpauno: We usually return `RMW_RET_ERROR` in this case.
In the case a specific rmw implementation doesn't support the feature, it has to still implement a function just returning `RMW_RET_UNSUPPORTED`.

rmw_implementation/src/functions.cpp
```diff
@@ -607,6 +625,10 @@ void prefetch_symbols(void)
   GET_SYMBOL(rmw_compare_gids_equal)
   GET_SYMBOL(rmw_service_server_is_available)
   GET_SYMBOL(rmw_set_log_severity)
+  GET_SYMBOL(rmw_get_publishers_info_by_topic)
+  rmw_reset_error();
```
jaisontj: This is because if the said function is not implemented by the implementation, then the GET_SYMBOL function replaces the previously set error and this prints warnings since rcutils_reset_error is not called.

rmw_implementation/src/functions.cpp
jaisontj: Hmm, maybe this implementation needs to be modified to catch `UNSUPPORTED` functions and return `RMW_RET_UNSUPPORTED` for those since this is an error caught [here](https://github.com/ros2/rmw_implementation/blob/6dd0abf8380378c6dc2e432aed2a73f049db9856/rmw_implementation/src/functions.cpp#L159) anyway and return `RMW_RET_ERROR` for the others. 

This is particularly important in throwing specific errors on the client layers (like [this](https://github.com/ros2/rclpy/pull/454/files#diff-80dda03110b5be606b823d6b0558b5c2R889)).

I guess, for now I'll change this back to `RMW_RET_ERROR`

ivanpauno: The problem, is that there are two ways of linking against `rmw`.

If there is more than one implementation, the code here in `rmw_implementation` is used, with all the dynamic loading stuff and symbol finding magic.

If you only have one implementation, the code is directly linked to the specific rmw implementation, ignoring `RMW_IMPLEMENTATION` environment variable at runtime (check https://github.com/ros2/rmw_implementation/blob/6dd0abf8380378c6dc2e432aed2a73f049db9856/rmw_implementation/CMakeLists.txt and https://github.com/ros2/rmw_implementation/blob/6dd0abf8380378c6dc2e432aed2a73f049db9856/rmw_implementation/rmw_implementation-extras.cmake.in).

If a specific rmw implementation doesn't implement a function, in the first case is detected at runtime and reported with an error code, and in the second linking would directly fail.

`RMW_RET_UNSUPPORTED` means that the specific rmw implementation that you're using doesn't support that feature, and that you can't use it as an user. It's not the idea to return that error if the symbol is not found, because that doesn't mean that the rmw implementation doesn't support that feature, but that is wrongly implemented and someone has to report a bug. IMO, `RMW_RET_ERROR` is the correct error code (and also consistent with the existing  code).

In any case, this is out of the scope of the PR. Please, open a follow-up issue if you think that this should be handled differently.

rmw_implementation/src/functions.cpp
ivanpauno: It's true that if more than one symbol is not found, a warning will be printed telling you that the error message was overridden. But again, it's out of the scope of the PR.
Please, open a follow-up issue, and we can apply the result of the discussion to the whole code consistently.

piraka9011:
@hidmic @ivanpauno  lets get this merged please :)

ivanpauno:
> @hidmic @ivanpauno lets get this merged please :)

I don't see any value in merging this until the comments in `rmw` and downstream PRs (`rmw_fastrtps`, etc) are addressed.
When that happens, I will merge them.

piraka9011:
> > @hidmic @ivanpauno lets get this merged please :)
> 
> I don't see any value in merging this until the comments in `rmw` and downstream PRs (`rmw_fastrtps`, etc) are addressed.
> When that happens, I will merge them.

Whoops, jumped the boat early on this one, though it was an earlier PR. My bad!!

PR #73, opened 2019-10-04
Add Python API for RMW implementation lookups

hidmic:
This pull request adds a basic Python API to lookup available RMW implementations in the system. Useful when parameterizing `pytests` by RMW implementation in pure Python packages i.e. no CMake templates available. 

Opening PR as a draft for preliminary discussion on whether this functionality should live here or in a separate package of its own.

### DIFF
--- a/rmw_implementation/rmw_implementation/__init__.py
+++ b/rmw_implementation/rmw_implementation/__init__.py
@@ -18,9 +18,9 @@ import ament_index_python
 
 
 def get_available_rmw_implementations():
-    """Return a list of all available RMW implementations as registered in the ament index."""
+    """Return the set of all available RMW implementations as registered in the ament index."""
     rmw_implementations = ament_index_python.get_resources('rmw_typesupport')
-    return [name for name in rmw_implementations if name != 'rmw_implementation']
+    return {name for name in rmw_implementations if name != 'rmw_implementation'}
 
 
 __all__ = [
--- a/rmw_implementation/rmw_implementation/__init__.py
+++ b/rmw_implementation/rmw_implementation/__init__.py
@@ -20,7 +20,7 @@ import ament_index_python
 def get_available_rmw_implementations():
     """Return a list of all available RMW implementations as registered in the ament index."""
     rmw_implementations = ament_index_python.get_resources('rmw_typesupport')
-    return [name for name, _ in rmw_implementations if name != 'rmw_implementation']
+    return [name for name in rmw_implementations if name != 'rmw_implementation']
 
 
 __all__ = [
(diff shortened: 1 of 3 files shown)


rmw_implementation/rmw_implementation/__init__.py
```diff
@@ -0,0 +1,28 @@
...
+
+"""Utilities to discover available RMW implementations."""
+
+import ament_index_python
+
+
+def get_available_rmw_implementations():
+    """Return a list of all available RMW implementations as registered in the ament index."""
+    rmw_implementations = ament_index_python.get_resources('rmw_typesupport')
+    return [name for name, _ in rmw_implementations if name != 'rmw_implementation']
```
dirk-thomas: Have you tried to run this code?

hidmic: Hmm, that's funny. I did, though not recently. Removed that extra `, _` in https://github.com/ros2/rmw_implementation/pull/73/commits/7a0309c2aadf8b29308feaf46f3093811b9b92ca.

rmw_implementation/rmw_implementation/__init__.py
```diff
@@ -0,0 +1,28 @@
...
+
+"""Utilities to discover available RMW implementations."""
+
+import ament_index_python
+
+
+def get_available_rmw_implementations():
+    """Return a list of all available RMW implementations as registered in the ament index."""
+    rmw_implementations = ament_index_python.get_resources('rmw_typesupport')
+    return [name for name in rmw_implementations if name != 'rmw_implementation']
```
dirk-thomas: How about returning a set here to not imply any order?

hidmic: Sounds perfectly reasonable, see https://github.com/ros2/rmw_implementation/pull/73/commits/5a7690aebcd12e6f920a65b474076eac48ef1350.

hidmic:
Getting this one in to re-run CI on https://github.com/ros2/ros2cli/pull/304.

wjwwood:
We should consider moving this to a different package. This change causes all of the client libraries to depend on Python indirectly, which probably isn't ideal... If it were a separate package, them packages like rcl or rclcpp could still depend on and use this python code, but they could do so via a build or buildtool depend (I'm pretty sure it's not being used by them at runtime). But since this is an exec depend and `rmw_implementation` is most often an exec depend (perhaps transitively) it causes a runtime dependency on `ament_index_python` and therefore Python itself.

dirk-thomas:
@wjwwood Can you please create a separate ticket for this. Otherwise the chance is high that your comment on a closed ticket will get lost.

PR #74, opened 2019-10-04
update signature for added pub/sub options

wjwwood:
Connects to ros2/rclcpp#882


mjcarroll:
`Epr` is expected to fail here because the underlying `rmw` API has changed.

PR #77, opened 2019-12-18
Use one participant per context API changes

ivanpauno:
Minimal renaming as part of https://github.com/ros2/rmw/issues/183.


hidmic:
Still waiting for CI

PR #81, opened 2020-02-11
Added conditional inclusion of Threads

pablogs9:
We are working on micro-ROS crosscompilation and under certain embedded environments, Threads package is not available.

It seems that Threads is not used along rwm_implementation, so it is proposed to conditionally include it when `BUILD_TESTING` is enabled.

### DIFF
--- a/rmw_implementation/rmw_implementation-extras.cmake.in
+++ b/rmw_implementation/rmw_implementation-extras.cmake.in
@@ -64,5 +64,6 @@ else()
   # since this code is already part of a find_package call of that package
 endif()
 
-find_package(Threads REQUIRED)
+find_package(Threads)
+
 list(APPEND rmw_implementation_LIBRARIES "${CMAKE_THREAD_LIBS_INIT}")
--- a/rmw_implementation/rmw_implementation-extras.cmake.in
+++ b/rmw_implementation/rmw_implementation-extras.cmake.in
@@ -64,6 +64,12 @@ else()
   # since this code is already part of a find_package call of that package
 endif()
 
-find_package(Threads)
+if(CMAKE_CROSSCOMPILING)
+  find_package(Threads QUIET)
+else()
+  find_package(Threads REQUIRED)
+endif()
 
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Please elaborate why it should be exported when building tests.

pablogs9:
In fact, Threads seems not used anywhere in this package. Are you ok with removing this package?

dirk-thomas:
I doubt this exists for no reason. Have you tried building downstream packages from source with it being removed? With multiple RMW as well as with only a single RWM?

pablogs9:
Well, we have found that some other packages depends on this Threads package included in the rmw_implementation. What do you think about setting it as not required?

```
find_package(Threads)
```

dirk-thomas:
> What do you think about setting it as not required?

Just making the change so it "works" without knowing why isn't sufficient. It needs to be clear why a dependency is needed or not.

pablogs9:
We have seen that the inclusion of pthread library was on https://github.com/ros2/rmw_implementation/pull/53 due to some crosscompilation issues. 

We are on the same situation because some micro-ROS target RTOS dont have pthread support. So, we propose to mark it as not required or using a flag that disables this requirement.

What approach is the best in your opinion?

Thanks.

dirk-thomas:
> What approach is the best in your opinion?

To determine when it is required and making that explicit.

Just making it not required will fail in the cases where it actually is.

pablogs9:
On some crosscompiling environments (e.g. embedded RTOS such as Zephyr or FreeRTOS) there is no Threads support so, what about enabling this requirement to be quiet in crosscompiling environments?

dirk-thomas:
You are only trying to workaround the question here. Please first clarify why exactly it is there in the first place. Assuming it can not be removed why is it needed? And second: why is it ok to omit it in your case of cross compilation? Obviously it was added in the past because it was missing in some cases - including cross compilation.

dirk-thomas:
@pablogs9 Friendly ping.

pablogs9:
We found a workaround. Sorry for not closing!

Issue #83, opened 2020-02-25
Move rmw_implementation python code to a separate package


ahcorde:
`rmw_implementation` has a python funtion called `get_available_rmw_implementations`. This function is being used in ros2cli in some of these packages' test

I have created this issue according with this PR https://github.com/ros2/rmw_implementation/pull/73#issuecomment-566804230

PR #84, opened 2020-02-25
Splitted the package in rmw_implementation and rmw_implementation_python

ahcorde:
According with this issue https://github.com/ros2/rmw_implementation/issues/83 I spplited the package in two:

 - rmw_implementation
 - rmw_implementation_python


rmw_implementation_python/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
+cmake_minimum_required(VERSION 3.5)
+
+project(rmw_implementation_python)
+
+find_package(ament_cmake REQUIRED)
+find_package(ament_cmake_python REQUIRED)
+
+ament_python_install_package(${PROJECT_NAME})
```
wjwwood: This could probably be made into a pure Python package with a `setup.py` rather than a `CMakeLists.txt`, but this also works as-is. I just wanted to make sure you were aware it was possible.

wjwwood:
This satisfies what we need for the quality work (removing Python as an exec depend).

There are some other use cases which would benefit from having this out of the repository (building from source C++ only), for example. Either by having this new package in a separate repository, or by eliminating this new package and just putting the logic in something like `launch_ros_testing` instead.

Anyways, like I said, LGTM and it meets the needs of our work. I just wanted to point out the other options we could have taken or that someone else could take in the future.

rmw_implementation_python/CMakeLists.txt
dirk-thomas: This should really not be a Python package. So :-1: from me to merge this as-is.

wjwwood: > This should really not be a Python package.

It isn't a Python package right now? Did you mean it should really be a Python package?

wjwwood:
> I don't see enough value to create a package for a single one-line function which doesn't even contain any valuable logic.

I don't really agree with this (I think there's a good chance we will need this in more places in the future, at which point we'll need to make into a package again).

However, I don't really care if it is just copy-pasted to the handle full of cases where it is currently used. If in the future we need it in more places we can easily make the package at the point.

rmw_implementation_python/CMakeLists.txt
dirk-thomas: That's indeed what I meant.

ahcorde:
Removed python code and using ament_index_python.get_resources()

 - rmw_implementation https://github.com/ros2/rmw_implementation/pull/85
 - ros2cli https://github.com/ros2/ros2cli/pull/460

PR #86, opened 2020-03-04
added rosidl_runtime_c depencency

ahcorde:
This PR is related to the changes introduced in this PR https://github.com/ros2/rosidl/pull/442. The full process can be followed here https://github.com/ros2/rosidl/issues/443

### DIFF
--- a/rmw_implementation/package.xml
+++ b/rmw_implementation/package.xml
@@ -22,6 +22,7 @@
   <build_depend>rmw_cyclonedds_cpp</build_depend>
   <build_depend>rmw_fastrtps_cpp</build_depend>
   <build_depend>rmw_implementation_cmake</build_depend>
+  <build_depend>rosidl_runtime_c</build_depend>
   <!-- end of group dependencies added for bloom -->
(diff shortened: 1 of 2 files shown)


rmw_implementation/package.xml
```diff
@@ -22,6 +22,7 @@
   <build_depend>rmw_cyclonedds_cpp</build_depend>
   <build_depend>rmw_fastrtps_cpp</build_depend>
   <build_depend>rmw_implementation_cmake</build_depend>
+  <build_depend>rosidl_runtime_c</build_depend>
```
wjwwood: Is this package part of the `rmw_implementation_packages` group?

ahcorde: I think it's part of the `rosidl_default_runtime`

dirk-thomas:
@ahcorde Can you please describe *why* the dependency is being added?

ahcorde:
my bad, the dependency should be changed here https://github.com/ros2/rmw/pull/198

PR #87, opened 2020-03-13
Removed poco dependency

ahcorde:
As part of the effort to remove Poco dependency described in this issue https://github.com/ros2/ros2/issues/867 this PR makes use of the recent changes in rcutils https://github.com/ros2/rcutils/pull/215 to load, unload and get symbols from shared libraries instead of using Poco.

### DIFF
--- a/rmw_implementation/src/functions.cpp
+++ b/rmw_implementation/src/functions.cpp
@@ -69,7 +69,7 @@ get_library()
       lib = std::make_shared<rcpputils::SharedLibrary>(library_path.c_str());
     } catch (const std::runtime_error & e) {
       RMW_SET_ERROR_MSG(
-        ("failed to load shared library of rmw implementation." + library_path +
+        ("failed to load shared library of rmw implementation: " + library_path +
         " Exception: " + std::string(e.what())).c_str());
       return nullptr;
     } catch (const std::bad_alloc & e) {
--- a/rmw_implementation/src/functions.cpp
+++ b/rmw_implementation/src/functions.cpp
@@ -88,6 +88,7 @@ get_symbol(const char * symbol_name)
   std::shared_ptr<rcpputils::SharedLibrary> lib = get_library();
 
   if (!lib) {
+    // error message set by get_library()
     return nullptr;
   }
 
--- a/rmw_implementation/src/functions.cpp
+++ b/rmw_implementation/src/functions.cpp
@@ -69,8 +69,8 @@ get_library()
       lib = std::make_shared<rcpputils::SharedLibrary>(library_path.c_str());
     } catch (const std::runtime_error & e) {
       RMW_SET_ERROR_MSG(
-        ("failed to allocate memory " + library_path + ": " + std::string(
-          e.what())).c_str());
+        ("failed to load shared library of rmw implementation." + library_path +
+        " Exception: " + std::string(e.what())).c_str());
       return nullptr;
     } catch (const std::bad_alloc & e) {
       RMW_SET_ERROR_MSG(
--- a/rmw_implementation/src/functions.cpp
+++ b/rmw_implementation/src/functions.cpp
@@ -90,7 +90,8 @@ get_symbol(const char * symbol_name)
     rcutils_allocator_t allocator = rcutils_get_default_allocator();
     char * msg = rcutils_format_string(
       allocator,
-      "failed to resolve symbol '%s' in shared library", symbol_name);
+      "failed to resolve symbol '%s' in shared library '%s'", symbol_name,
+      lib->get_library_path().c_str());
     if (msg) {
       RMW_SET_ERROR_MSG(msg);
       allocator.deallocate(msg, allocator.state);
--- a/rmw_implementation/src/functions.cpp
+++ b/rmw_implementation/src/functions.cpp
@@ -13,15 +13,15 @@
 // limitations under the License.
 
 #include <cstddef>
+#include <stdexcept>
 
 #include <string>
 
 #include "rcutils/allocator.h"
 #include "rcutils/format_string.h"
 #include "rcutils/get_env.h"
+#include "rcutils/shared_library.h"
 #include "rcutils/types/string_array.h"
-
-#include "Poco/SharedLibrary.h"
 
 #include "rcpputils/find_library.hpp"
 #include "rmw/error_handling.h"
@@ -46,11 +46,22 @@ std::string get_env_var(const char * env_var)
   return value ? value : "";
 }
 
-Poco::SharedLibrary *
+rcutils_shared_library_t *
 get_library()
 {
-  static Poco::SharedLibrary * lib = nullptr;
+  static rcutils_shared_library_t * lib = nullptr;
+  rcutils_allocator_t allocator = rcutils_get_default_allocator();
+
   if (!lib) {
+    lib = static_cast<rcutils_shared_library_t *>(allocator.allocate(
+            sizeof(rcutils_shared_library_t), allocator.state));
+    if (!lib) {
+      RMW_SET_ERROR_MSG("failed to allocate memory");
+      return nullptr;
+    }
+
+    *lib = rcutils_get_zero_initialized_shared_library();
+
     std::string env_var = get_env_var("RMW_IMPLEMENTATION");
     if (env_var.empty()) {
       env_var = STRINGIFY(DEFAULT_RMW_IMPLEMENTATION);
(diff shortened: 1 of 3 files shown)


rmw_implementation/src/functions.cpp
```diff
@@ -46,11 +46,22 @@ std::string get_env_var(const char * env_var)
...
+rcutils_shared_library_t *
 get_library()
 {
-  static Poco::SharedLibrary * lib = nullptr;
+  static rcutils_shared_library_t * lib = nullptr;
+  rcutils_allocator_t allocator = rcutils_get_default_allocator();
+
   if (!lib) {
+    lib = static_cast<rcutils_shared_library_t *>(allocator.allocate(
+            sizeof(rcutils_shared_library_t), allocator.state));
```
wjwwood: Same kind of questions/comments as the other pr:

- weird indent
- shouldn't this be down where the `new` used to be?

wjwwood: Also, make sure this is not leaked when this function fails in other cases below...

rmw_implementation/src/functions.cpp
```diff
@@ -61,17 +72,9 @@ get_library()
...
-        e.displayText()).c_str());
-      return nullptr;
-    } catch (...) {
-      RMW_SET_ERROR_MSG("failed to load shared library of rmw implementation");
-      RMW_SET_ERROR_MSG(
-        ("failed to load shared library of rmw implementation: " + library_path).c_str());
+    rcutils_ret_t ret = rcutils_load_shared_library(lib, library_path.c_str());
+    if (ret != RCUTILS_RET_OK) {
+      RMW_SET_ERROR_MSG(("Cannot open library " + library_path).c_str());
       return nullptr;
```
wjwwood: Like if you return here, isn't `lib` going to be leaked?

rmw_implementation/src/functions.cpp
```diff
@@ -61,17 +72,9 @@ get_library()
         ("failed to find shared library of rmw implementation. Searched " + env_var).c_str());
       return nullptr;
     }
-    try {
-      lib = new Poco::SharedLibrary(library_path);
-    } catch (Poco::LibraryLoadException & e) {
-      RMW_SET_ERROR_MSG(
-        ("failed to load shared library of rmw implementation. Exception: " +
-        e.displayText()).c_str());
-      return nullptr;
```
wjwwood: See this version of the dynamic loading actually catches exceptions and uses error states properly and returns error codes, making it very C-like. Perhaps the other similar functions in `rosidl_typesupport_c/cpp` should be updated to look and act like this one...

rmw_implementation/src/functions.cpp
```diff
@@ -81,16 +84,19 @@ get_library()
 void *
 get_symbol(const char * symbol_name)
 {
-  Poco::SharedLibrary * lib = get_library();
+  rcutils_shared_library_t * lib = get_library();
```
wjwwood: This seems weird to me that it is making a new shared library handle each time. I guess it's essentially a noop for dlopen and the related functions, but I wonder if poco was making this into a single for us... Perhaps this should be shard resource rather than being loaded each time `get_symbol` is called?

Also, when does this shared library get unloaded? never?

ahcorde: Library is declared here as `static` 
https://github.com/ros2/rmw_implementation/blob/da6ca0af16b57469fa88df0566eababe8b8fb947/rmw_implementation/src/functions.cpp#L52-L55

Then it's checking each time if the library is already load.

This code loads `rmw_X_cpp.so` library I think this library should be released when :
 - the node is killed?
 - ?

rmw_implementation/src/functions.cpp
ahcorde: used `new`

rmw_implementation/src/functions.cpp
ahcorde: used `delete` when a failure happened

rmw_implementation/src/functions.cpp
```diff
@@ -61,17 +64,13 @@ get_library()
         ("failed to find shared library of rmw implementation. Searched " + env_var).c_str());
       return nullptr;
     }
+
     try {
-      lib = new Poco::SharedLibrary(library_path);
-    } catch (Poco::LibraryLoadException & e) {
+      lib = std::make_shared<rcpputils::SharedLibrary>(library_path.c_str());
+    } catch (const std::runtime_error & e) {
```
dirk-thomas: Same as https://github.com/ros2/rosidl_typesupport/pull/59/files#r396595712.

Same below.

rmw_implementation/src/functions.cpp
```diff
@@ -81,16 +80,17 @@ get_library()
...
     return nullptr;
   }
-  if (!lib->hasSymbol(symbol_name)) {
+
+  if (!lib->has_symbol(symbol_name)) {
     rcutils_allocator_t allocator = rcutils_get_default_allocator();
     char * msg = rcutils_format_string(
       allocator,
-      "failed to resolve symbol '%s' in shared library '%s'", symbol_name, lib->getPath().c_str());
+      "failed to resolve symbol '%s' in shared library", symbol_name);
```
dirk-thomas: Why should the library path be removed from the error message?

ahcorde:
rcpputils PR is merged, can you review this ? @wjwwood or @dirk-thomas

rmw_implementation/src/functions.cpp
```diff
@@ -81,16 +85,18 @@ get_library()
 void *
 get_symbol(const char * symbol_name)
 {
-  Poco::SharedLibrary * lib = get_library();
+  std::shared_ptr<rcpputils::SharedLibrary> lib = get_library();
+
   if (!lib) {
-    // error message set by get_library()
```
dirk-thomas, line 86: Why was this comment removed? It seems the function still sets an error message.

rmw_implementation/src/functions.cpp
```diff
@@ -61,17 +64,18 @@ get_library()
...
-        e.displayText()).c_str());
+        ("failed to allocate memory " + library_path + ": " + std::string(
+          e.what())).c_str());
       return nullptr;
-    } catch (...) {
-      RMW_SET_ERROR_MSG("failed to load shared library of rmw implementation");
+    } catch (const std::bad_alloc & e) {
       RMW_SET_ERROR_MSG(
-        ("failed to load shared library of rmw implementation: " + library_path).c_str());
+        ("Cannot open library " + library_path + ": " + std::string(
```
dirk-thomas, line 77: Why change the static part of the error message? The new string doesn't mention "rmw implementation" at all. If there is no reason to change please keep the static part of the message as it was before.

ahcorde: Restored

rmw_implementation/src/functions.cpp
```diff
@@ -61,17 +64,18 @@ get_library()
...
+
     try {
-      lib = new Poco::SharedLibrary(library_path);
-    } catch (Poco::LibraryLoadException & e) {
+      lib = std::make_shared<rcpputils::SharedLibrary>(library_path.c_str());
+    } catch (const std::runtime_error & e) {
       RMW_SET_ERROR_MSG(
-        ("failed to load shared library of rmw implementation. Exception: " +
-        e.displayText()).c_str());
+        ("failed to allocate memory " + library_path + ": " + std::string(
```
dirk-thomas, line 72: Same here - also this case is not about memory allocation at all.

rmw_implementation/src/functions.cpp
```diff
@@ -61,17 +64,18 @@ get_library()
...
+
     try {
-      lib = new Poco::SharedLibrary(library_path);
-    } catch (Poco::LibraryLoadException & e) {
+      lib = std::make_shared<rcpputils::SharedLibrary>(library_path.c_str());
+    } catch (const std::runtime_error & e) {
       RMW_SET_ERROR_MSG(
-        ("failed to load shared library of rmw implementation. Exception: " +
-        e.displayText()).c_str());
+        ("failed to load shared library of rmw implementation." + library_path +
```
dirk-thomas, line 72: Nitpick: the message has a period between the static text and the library path and no space - looks weird to me.

ahcorde:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9927/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5537/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8075/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9773/)

ahcorde:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9929/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5539/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8077/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9775/)

PR #106, opened 2020-06-19
Add test_rmw_implementation package.

hidmic:
Precisely what the title says. To serve as a basis for upcoming generic RMW API tests. Includes dummy init/shutdown test.

### DIFF
--- a/test_rmw_implementation/test/test_init_shutdown.cpp
+++ b/test_rmw_implementation/test/test_init_shutdown.cpp
@@ -30,7 +30,7 @@
 
 class CLASSNAME (TestInitShutdown, RMW_IMPLEMENTATION) : public ::testing::Test {};
 
-TEST_F(TestInitShutdown, init_shutdown) {
+TEST_F(CLASSNAME (TestInitShutdown, RMW_IMPLEMENTATION), init_shutdown) {
   rmw_context_t context = rmw_get_zero_initialized_context();
   rmw_init_options_t options = rmw_get_zero_initialized_init_options();
   rmw_ret_t ret = rmw_init_options_init(&options, rcutils_get_default_allocator());
--- a/test_rmw_implementation/test/test_init_shutdown.cpp
+++ b/test_rmw_implementation/test/test_init_shutdown.cpp
@@ -35,6 +35,11 @@ TEST(TestInitShutdown, init_shutdown) {
 
   ret = rmw_init(&options, &context);
   ASSERT_EQ(RMW_RET_OK, ret) << rcutils_get_error_string().str;
+  OSRF_TESTING_TOOLS_CPP_SCOPE_EXIT(
+  {
+    rmw_ret_t ret = rmw_context_fini(&context);
+    EXPECT_EQ(RMW_RET_OK, ret) << rcutils_get_error_string().str;
+  });
 
   ret = rmw_shutdown(&context);
   EXPECT_EQ(RMW_RET_OK, ret) << rcutils_get_error_string().str;
--- a/test_rmw_implementation/test/test_init_shutdown.cpp
+++ b/test_rmw_implementation/test/test_init_shutdown.cpp
@@ -21,8 +21,16 @@
 
 #include "rmw/rmw.h"
 
+#ifdef RMW_IMPLEMENTATION
+# define CLASSNAME_(NAME, SUFFIX) NAME ## __ ## SUFFIX
+# define CLASSNAME(NAME, SUFFIX) CLASSNAME_(NAME, SUFFIX)
+#else
+# define CLASSNAME(NAME, SUFFIX) NAME
+#endif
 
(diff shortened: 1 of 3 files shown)


hidmic:
Interestingly enough, this test as-is fails for `rmw_cyclonedds_cpp`. It cannot handle the `RMW_DEFAULT_DOMAIN_ID` and fails with:

```sh
[rmw_cyclonedds_cpp]: rmw_create_node: domain id out of range
```

It appears we didn't notice before because `rcl_get_default_domain_id()` implementation will always force the domain id to 0 if no `ROS_DOMAIN_ID` env var is present or it if contains anything other than a number (because of [`strtoul` use](https://github.com/ros2/rcl/blob/ffcfda18db7aa9ad7c01ebdd26ae7bdd73c8206a/rcl/src/rcl/domain_id.c#L44)).

Shall we change `RMW_DEFAULT_DOMAIN_ID` to reflect that?

test_rmw_implementation/test/test_init_shutdown.cpp
```diff
@@ -0,0 +1,46 @@
...
+
+#include "osrf_testing_tools_cpp/scope_exit.hpp"
+
+#include "rcutils/allocator.h"
+#include "rcutils/error_handling.h"
+
+#include "rmw/rmw.h"
+
+
+TEST(TestInitShutdown, init_shutdown) {
```
brawner, line 25: Do these test class names need to be modified for their rmw_implementation?
``
TEST(CLASSNAME(TestInitShutdown, RMW_IMPLEMENTATION), init_shutdown)
```

Comment (hidmic) [reply test_rmw_implementation/test/test_init_shutdown.cpp:25]: Yes! Good catch!

Comment (wjwwood) [reply test_rmw_implementation/test/test_init_shutdown.cpp:25]: Was just typing this haha

Comment (hidmic) [reply test_rmw_implementation/test/test_init_shutdown.cpp]: See https://github.com/ros2/rmw_implementation/pull/106/commits/228f34bbbffba532cc91a5edba2b20d107e4987c.

**Comment by Karsten1987** [comment]
just my 2cts, but why do we need a dedicated package for this? Can't this be part of system tests?

**Comment by hidmic** [comment]
> just my 2cts, but why do we need a dedicated package for this? Can't this be part of system tests?

This package is not meant to include system tests, but (ideally) unit tests for every portion of the documented `rmw` API. 

While much of `rmw`'s functionality, as well as `rcl`'s, `rclpy`'s and `rclcpp`'s, is indirectly covered by integration and system tests up the stack, many corner cases cannot be easily checked.

File: test_rmw_implementation/test/test_init_shutdown.cpp [hunk 2]
```diff
@@ -0,0 +1,54 @@
...
+#ifdef RMW_IMPLEMENTATION
+# define CLASSNAME_(NAME, SUFFIX) NAME ## __ ## SUFFIX
+# define CLASSNAME(NAME, SUFFIX) CLASSNAME_(NAME, SUFFIX)
+#else
+# define CLASSNAME(NAME, SUFFIX) NAME
+#endif
+
+class CLASSNAME (TestInitShutdown, RMW_IMPLEMENTATION) : public ::testing::Test {};
+
+TEST_F(TestInitShutdown, init_shutdown) {
```
wjwwood, line 33: This also needs to use the `CLASSNAME` macro, no?

hidmic: Oof, I had a brainfart.

hidmic:
> I think having a way to let the middleware pick the default is useful, so maybe we should change it so it is honored.

About this. Seems like neither `rmw_fastrtps_cpp` nor `rmw_connext_cpp` are doing anything with it. They just pass along whatever ID it came their way. `rmw_cyclonedds_cpp` is more opinionated and has specific code to reject such default.

hidmic:
CI up to `rcl` and `test_rmw_implementation` against all RMW implementations:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11229/) (canary build)

hidmic:
I was missing one last bit! See https://github.com/ros2/rmw_implementation/pull/106/commits/7fde494797d323059592d3b2f8f8c4c916664c4b.

dirk-thomas:
@hidmic please follow up with releases as necessary to make this job pass again: http://build.ros2.org/job/Rdev__rmw_implementation__ubuntu_focal_amd64/2/testReport/

sloretz:
Looks like this PR caused two CI jobs to fail last night. I think the fix is just to add a `<test_depend>` for `ament_cmake_gtest`. Opening a PR to try that now.

* http://build.ros2.org/job/Fci__nightly-connext_ubuntu_focal_amd64/148/console
* http://build.ros2.org/job/Fci__nightly-cyclonedds_ubuntu_focal_amd64/150/console

PR #107, opened 2020-06-22
Complete init/shutdown API test coverage.

hidmic:
Precisely what the title says. Follow-up after #106.

### DIFF
--- a/test_rmw_implementation/test/test_init_shutdown.cpp
+++ b/test_rmw_implementation/test/test_init_shutdown.cpp
@@ -28,20 +28,94 @@
 # define CLASSNAME(NAME, SUFFIX) NAME
 #endif
 
-class CLASSNAME (TestInitShutdown, RMW_IMPLEMENTATION) : public ::testing::Test {};
+class CLASSNAME (TestInitShutdown, RMW_IMPLEMENTATION) : public ::testing::Test
+{
+protected:
+  void SetUp() override
+  {
+    options = rmw_get_zero_initialized_init_options();
+    rmw_ret_t ret = rmw_init_options_init(&options, rcutils_get_default_allocator());
+    ASSERT_EQ(RMW_RET_OK, ret) << rcutils_get_error_string().str;
+  }
 
-TEST_F(CLASSNAME(TestInitShutdown, RMW_IMPLEMENTATION), init_shutdown) {
-  rmw_context_t context = rmw_get_zero_initialized_context();
-  rmw_init_options_t options = rmw_get_zero_initialized_init_options();
-  rmw_ret_t ret = rmw_init_options_init(&options, rcutils_get_default_allocator());
-  ASSERT_EQ(RMW_RET_OK, ret) << rcutils_get_error_string().str;
-  OSRF_TESTING_TOOLS_CPP_SCOPE_EXIT(
+  void TearDown() override
   {
     rmw_ret_t ret = rmw_init_options_fini(&options);
     EXPECT_EQ(RMW_RET_OK, ret) << rcutils_get_error_string().str;
+  }
+
+  rmw_init_options_t options;
+};
+
+TEST_F(CLASSNAME(TestInitShutdown, RMW_IMPLEMENTATION), init_with_bad_arguments) {
+  rmw_context_t context = rmw_get_zero_initialized_context();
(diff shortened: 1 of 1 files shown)


hidmic:
These tests need https://github.com/ros2/rmw/pull/242 and connected PRs to pass.

test_rmw_implementation/test/test_init_shutdown.cpp
```diff
@@ -28,20 +28,94 @@
...
+
+TEST_F(CLASSNAME(TestInitShutdown, RMW_IMPLEMENTATION), shutdown_with_bad_arguments) {
+  EXPECT_EQ(RMW_RET_INVALID_ARGUMENT, rmw_shutdown(nullptr));
+  rcutils_reset_error();
 
-TEST_F(CLASSNAME(TestInitShutdown, RMW_IMPLEMENTATION), init_shutdown) {
   rmw_context_t context = rmw_get_zero_initialized_context();
-  rmw_init_options_t options = rmw_get_zero_initialized_init_options();
-  rmw_ret_t ret = rmw_init_options_init(&options, rcutils_get_default_allocator());
+  EXPECT_NE(RMW_RET_OK, rmw_shutdown(&context));
```
brawner, line 71: Is there a specific error you are expecting? RMW_RET_ERROR?

hidmic: Based on [`rmw_shutdown()` API contract](https://github.com/ros2/rmw/blob/b05f7eacd0e3e4209ffdcac40f3d353e932511b3/rmw/include/rmw/init.h#L86-L106), an RMW implementation could equally return `RMW_RET_INVALID_ARGUMENT` or `RMW_RET_INCORRECT_IMPLEMENTATION`. That's why I don't check for it.

brawner: Since you already test the incorrect_implementation error below, wouldn't it make more sense to test the invalid_argument error here? Set context.implementation_identifier to a known valid identifier, then check that it returns invalid argument?

hidmic: Hmm, problem is that we get into bad initialization territory. Here I deal with the four (4) known context states: zero-initialized, initialized and not shutdown, initialized and shutdown, finalized. Anything outside that is UB. So if we want to be specific about the assertion we have to further clarify API contracts i.e. what it means for a context to be invalid.

brawner: If you don't want to monkey around with the identifier field to force the invalid argument, then I recommend to at least narrow down the expect statement to only two errors you expect. Something like:

```
EXPECT_TRUE((ret == RMW_RET_INVALID_ARGUMENT || 
             ret == RMW_RET_INCORRECT_IMPLEMENTATION))
```

hidmic:
CI up to `test_rmw_implementation` against all RMW implementations:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11240/) (canary build)

test_rmw_implementation/test/test_init_shutdown.cpp
hidmic: Sure, we can do that. See  https://github.com/ros2/rmw_implementation/pull/107/commits/9bff7f27014246338bf1fe84322993a574db9f3a

hidmic:
Hold on, as tests may change with https://github.com/ros2/rmw/pull/243.

hidmic:
Ready for re-review https://github.com/ros2/rmw_implementation/pull/107/commits/bc7aa7f8c34971b57ba623ddaa7ad9be8d078e22.

brawner:
Does this need to be retested on CI?

hidmic:
> Does this need to be retested on CI?

More than that. RMW implementations have to be released into `rolling` after connected PRs have to be merged to get that PR job passing.

hidmic:
CI up to `test_rmw_implementation` against all RMW implementations:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11320/)

hidmic:
Same as in https://github.com/ros2/rmw_implementation/pull/108, I'll hold off merging this until I've re-released RMWs so that rolling PR jobs pass.

hidmic:
@ros-pull-request-builder retest this please

hidmic:
Alright, all's green! Going in.

PR #109, opened 2020-06-24
Add dependency on ament_cmake_gtest

sloretz:
https://github.com/ros2/rmw_implementation/pull/106#issuecomment-648489712

@hidmic FYI

### DIFF
--- a/test_rmw_implementation/package.xml
+++ b/test_rmw_implementation/package.xml
@@ -8,15 +8,15 @@
   <license>Apache License 2.0</license>
 
   <buildtool_depend>ament_cmake</buildtool_depend>
-  <buildtool_depend>ament_cmake_gtest</buildtool_depend>
-  <buildtool_depend>rmw_implementation_cmake</buildtool_depend>
 
+  <test_depend>ament_cmake_gtest</test_depend>
   <test_depend>ament_lint_auto</test_depend>
   <test_depend>ament_lint_common</test_depend>
   <test_depend>osrf_testing_tools_cpp</test_depend>
   <test_depend>rcutils</test_depend>
   <test_depend>rmw</test_depend>
   <test_depend>rmw_implementation</test_depend>
+  <test_depend>rmw_implementation_cmake</test_depend>
 
   <export>
     <build_type>ament_cmake</build_type>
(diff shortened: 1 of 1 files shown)


sloretz:
CI (building up to and testing only `test_rmw_implementation`)

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11274/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6530/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9231/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11179/)

dirk-thomas:
Shouldn't it be a `test_depend` instead?

sloretz:
> Shouldn't it be a test_depend instead?

I think either works. I picked `buildtool_depend` to match `<buildtool_depend>rmw_implementation_cmake</buildtool_depend>`  which could also be a `test_depend` for the same reason.

sloretz:
> Shouldn't it be a test_depend instead?

Actually, it brings up an interesting case with the `package.xml` when cross-compiling. Does `<test_depend>` mean the dependency is needed on the target machine or the buildtool machine in order to run tests? Maybe there should be a `<buildtool_test_depend>` for dependencies needed only on the buildtool machine to run tests..

sloretz:
> Maybe there should be a <buildtool_test_depend> for dependencies needed only on the buildtool machine to run tests..

Ah nevermind, [REP 149 already covers this](https://ros.org/reps/rep-0149.html#test-depend-multiple):

> When building with testing enabled, the <test_depend> packages are available for configuring and building the tests as well as running them. Generated Debian packages are built without the unit tests or their dependencies.

sloretz:
> Shouldn't it be a test_depend instead?

@dirk-thomas `rmw_implementation_cmake` and `ament_cmake_gtest` made `test_depend` in f7e0563

sloretz:
CI (building and testing `test_rmw_implementation only)
* Linux only as I'll assume a package.xml only change won't affect other platforms [Build Status](https://ci.ros2.org/job/ci_linux/11275/)

hidmic:
Uh, I thought I had added the dependency. Oh well, thanks for the fix @sloretz !!

PR #110, opened 2020-06-25
Update Quality Declaration to QL3.

hidmic:


hidmic:
On a related note, now that this package isn't alone in the repository, I think QD, LICENSE and README.md should go into the package.

PR #111, opened 2020-06-26
Move the QD into the rmw_implementation subdirectory.

clalancette:
That is more appropriate, as it only applies to that package.


clalancette:
> @clalancette just moving the file isn't enough -- there're plenty of local links in this declaration.

Yep, good call.  I think I've fixed them all in 90a690e.

> Also, I think the README.md should also be moved into the package. WDYT?

Agreed.  Also moved in 90a690e.

PR #115, opened 2020-07-20
Add publisher API test coverage

hidmic:
Depends on https://github.com/ros2/rmw/pull/252.

### DIFF
--- a/test_rmw_implementation/test/test_publisher.cpp
+++ b/test_rmw_implementation/test/test_publisher.cpp
@@ -42,8 +42,8 @@ protected:
     context = rmw_get_zero_initialized_context();
     ret = rmw_init(&init_options, &context);
     ASSERT_EQ(RMW_RET_OK, ret) << rcutils_get_error_string().str;
-    constexpr const char * node_name = "my_test_node";
-    constexpr const char * node_namespace = "/my_test_ns";
+    constexpr char node_name[] = "my_test_node";
+    constexpr char node_namespace[] = "/my_test_ns";
     node = rmw_create_node(&context, node_name, node_namespace);
     ASSERT_NE(nullptr, node) << rcutils_get_error_string().str;
   }
@@ -67,7 +67,7 @@ protected:
 
 TEST_F(CLASSNAME(TestPublisher, RMW_IMPLEMENTATION), create_and_destroy) {
   rmw_publisher_options_t options = rmw_get_default_publisher_options();
-  constexpr const char * topic_name = "/test";
+  constexpr char topic_name[] = "/test";
   const rosidl_message_type_support_t * ts =
     ROSIDL_GET_MSG_TYPE_SUPPORT(test_msgs, msg, BasicTypes);
   rmw_publisher_t * pub =
@@ -79,7 +79,7 @@ TEST_F(CLASSNAME(TestPublisher, RMW_IMPLEMENTATION), create_and_destroy) {
 
 TEST_F(CLASSNAME(TestPublisher, RMW_IMPLEMENTATION), create_and_destroy_native) {
   rmw_publisher_options_t options = rmw_get_default_publisher_options();
-  constexpr const char * topic_name = "test";
+  constexpr char topic_name[] = "test";
   const rosidl_message_type_support_t * ts =
     ROSIDL_GET_MSG_TYPE_SUPPORT(test_msgs, msg, BasicTypes);
   rmw_qos_profile_t native_qos_profile = rmw_qos_profile_default;
@@ -93,7 +93,7 @@ TEST_F(CLASSNAME(TestPublisher, RMW_IMPLEMENTATION), create_and_destroy_native) 
 
 TEST_F(CLASSNAME(TestPublisher, RMW_IMPLEMENTATION), create_with_bad_arguments) {
   rmw_publisher_options_t options = rmw_get_default_publisher_options();
-  constexpr const char * topic_name = "/test";
+  constexpr char topic_name[] = "/test";
   const rosidl_message_type_support_t * ts =
     ROSIDL_GET_MSG_TYPE_SUPPORT(test_msgs, msg, BasicTypes);
   rmw_publisher_t * pub =
@@ -116,12 +116,12 @@ TEST_F(CLASSNAME(TestPublisher, RMW_IMPLEMENTATION), create_with_bad_arguments) 
   EXPECT_EQ(nullptr, pub);
   rmw_reset_error();
 
-  constexpr const char * topic_name_with_spaces = "/foo bar";
+  constexpr char topic_name_with_spaces[] = "/foo bar";
   pub = rmw_create_publisher(node, ts, topic_name_with_spaces, &rmw_qos_profile_default, &options);
   EXPECT_EQ(nullptr, pub);
   rmw_reset_error();
 
-  constexpr const char * relative_topic_name = "foo";
+  constexpr char relative_topic_name[] = "foo";
   pub = rmw_create_publisher(node, ts, relative_topic_name, &rmw_qos_profile_default, &options);
   EXPECT_EQ(nullptr, pub);
   rmw_reset_error();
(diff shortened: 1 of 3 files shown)


test_rmw_implementation/test/test_publisher.cpp
```diff
@@ -0,0 +1,245 @@
...
+
+  const char * implementation_identifier = node->implementation_identifier;
+  node->implementation_identifier = "not-an-rmw-implementation-identifier";
+  pub = rmw_create_publisher(node, ts, topic_name, &rmw_qos_profile_default, &options);
+  node->implementation_identifier = implementation_identifier;
+  EXPECT_EQ(nullptr, pub);
+  rmw_reset_error();
+
+  pub = rmw_create_publisher(node, ts, nullptr, &rmw_qos_profile_default, &options);
+  ASSERT_EQ(nullptr, pub);
```
Blast545, line 138: IMO those tests are independent, and if one fails doesn't affect the others, so the ASSERT statements can be changed to EXPECT ones.
```suggestion
  EXPECT_EQ(nullptr, pub);
```

test_rmw_implementation/test/test_publisher.cpp
```diff
@@ -0,0 +1,245 @@
...
+    ROSIDL_GET_MSG_TYPE_SUPPORT(test_msgs, msg, BasicTypes);
+  rmw_publisher_t * pub =
+    rmw_create_publisher(node, ts, topic_name, &rmw_qos_profile_system_default, &options);
+  ASSERT_NE(nullptr, pub) << rmw_get_error_string().str;
+  rmw_qos_profile_t qos_profile = rmw_qos_profile_unknown;
+  rmw_ret_t ret = rmw_publisher_get_actual_qos(pub, &qos_profile);
+  EXPECT_EQ(RMW_RET_OK, ret) << rmw_get_error_string().str;
+  EXPECT_NE(rmw_qos_profile_system_default.history, qos_profile.history);
+  EXPECT_NE(rmw_qos_profile_unknown.history, qos_profile.history);
+  EXPECT_NE(rmw_qos_profile_system_default.reliability, qos_profile.reliability);
```
Blast545, line 214: The publisher is created with `rmw_qos_profile_system_default` then why when reading the `qos_profile` do not match to this? It seems odd, mind asking a comment on this?

Blast545: I can't see the difference between this section vs the last test case `get_actual_qos`

hidmic: > The publisher is created with rmw_qos_profile_system_default then why when reading the qos_profile do not match to this? It seems odd, mind asking a comment on this?

`_SYSTEM_DEFAULT` are sentinel values, expect to be replaced by concrete policies. This test verifies that's the case in a generic way (i.e. can't test for a specific policy as it is implementation dependent).

> I can't see the difference between this section vs the last test case get_actual_qos

That other test uses concrete policies and so it expects them to be honored.

Just as an FYI, liveliness, lifespan and deadline QoS policies are less consistent in this regard and thus why they are currently not verified. I plan to come back to them at a later point in time.

test_rmw_implementation/test/test_publisher.cpp
hidmic: Good point. Also, there's a test above that should be two separate tests.

test_rmw_implementation/test/test_publisher.cpp
```diff
@@ -0,0 +1,245 @@
...
+  {
+    init_options = rmw_get_zero_initialized_init_options();
+    rmw_ret_t ret = rmw_init_options_init(&init_options, rcutils_get_default_allocator());
+    ASSERT_EQ(RMW_RET_OK, ret) << rcutils_get_error_string().str;
+    init_options.enclave = rcutils_strdup("/", rcutils_get_default_allocator());
+    ASSERT_STREQ("/", init_options.enclave);
+    context = rmw_get_zero_initialized_context();
+    ret = rmw_init(&init_options, &context);
+    ASSERT_EQ(RMW_RET_OK, ret) << rcutils_get_error_string().str;
+    const char * const node_name = "my_test_node";
```
brawner, line 45: Why not just constexpr? Here and below for string literals.

test_rmw_implementation/test/test_publisher.cpp
```diff
@@ -0,0 +1,245 @@
...
+    ret = rmw_init_options_fini(&init_options);
+    EXPECT_EQ(RMW_RET_OK, ret) << rmw_get_error_string().str;
+  }
+
+  rmw_init_options_t init_options;
+  rmw_context_t context;
+  rmw_node_t * node;
+};
+
+class CLASSNAME (TestPublisherUse, RMW_IMPLEMENTATION) :
```
brawner, line 68: Typically it's common to group classes with the tests yah? Should this class definition be moved to line 225?

test_rmw_implementation/test/test_publisher.cpp
```diff
@@ -0,0 +1,245 @@
...
+  pub = rmw_create_publisher(node, ts, topic_name, &rmw_qos_profile_default, &options);
+  node->implementation_identifier = implementation_identifier;
+  EXPECT_EQ(nullptr, pub);
+  rmw_reset_error();
+
+  pub = rmw_create_publisher(node, ts, nullptr, &rmw_qos_profile_default, &options);
+  ASSERT_EQ(nullptr, pub);
+  rmw_reset_error();
+
+  const char * const topic_name_with_spaces = "foo bar";
```
brawner, line 141: If relative topic names are also invalid, shouldn't this start with a leading '/'?

test_rmw_implementation/test/test_publisher.cpp
brawner: This test probably deserves a comment with this explanation.

test_rmw_implementation/test/test_publisher.cpp
hidmic: Good point.

test_rmw_implementation/test/test_publisher.cpp
hidmic: Sounds reasonable, moved.

test_rmw_implementation/test/test_publisher.cpp
hidmic: Sounds good. I expect the compiler to inline references to the global literal anyways.

test_rmw_implementation/test/test_publisher.cpp
```diff
@@ -42,8 +42,8 @@ class CLASSNAME (TestPublisher, RMW_IMPLEMENTATION) : public ::testing::Test
     context = rmw_get_zero_initialized_context();
     ret = rmw_init(&init_options, &context);
     ASSERT_EQ(RMW_RET_OK, ret) << rcutils_get_error_string().str;
-    const char * const node_name = "my_test_node";
-    const char * const node_namespace = "/my_test_ns";
+    constexpr const char * node_name = "my_test_node";
```
brawner, line 45:
```suggestion
    constexpr char node_name[] = "my_test_node";
```

hidmic: Sure, yeah. But for the record, that may force a stack allocation we don't need.

hidmic: For posterity, it turns out that, in exchange, using `constexpr` precludes string literals entirely and reduces somewhat the size of the binary. Thanks for the tip @brawner !

hidmic:
CI up `test_rmw_implementation` against all RMW implementations:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11630/)

brawner:
Why does the rpr build fail for connextdds? [Build Status](http://build.ros2.org/job/Rpr__rmw_implementation__ubuntu_focal_amd64/45/)

hidmic:
@brawner I have to merge related RMW PRs and release those repos to `rolling` before the PR job can pass.

hidmic:
@ros-pull-request-builder retest this please.

hidmic:
Alright, all green! Going in.

PR #116, opened 2020-07-22
Run tests for Fast-RTPS based implementations only.

hidmic:
Precisely what the title says. To speed up RMW layer specs refining and testing.

### DIFF
--- a/test_rmw_implementation/CMakeLists.txt
+++ b/test_rmw_implementation/CMakeLists.txt
@@ -23,39 +23,42 @@ if(BUILD_TESTING)
   find_package(rmw_implementation_cmake REQUIRED)
 
   macro(test_api)
-    find_package(${rmw_implementation} REQUIRED)
-    message(STATUS "Creating API tests for '${rmw_implementation}'")
-    set(rmw_implementation_env_var RMW_IMPLEMENTATION=${rmw_implementation})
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Why should we *not* test the other RMW impl?

hidmic:
> Why should we not test the other RMW impl?

My understanding is that we don't have the resources to keep updating all RMW implementations to comply with the RMW API specification as it evolves along with the tests. Thus, it's unlikely that without changes those implementations will pass these tests. 

I'm opening this for debate so that, both if we forward with this or we don't, we are all aware of the implications.

chapulina:
> we don't have the resources

I'm looking into this, let's not merge this PR yet :crossed_fingers:

dirk-thomas:
> My understanding is that we don't have the resources to keep updating all RMW implementations to comply with the RMW API specification as it evolves along with the tests. Thus, it's unlikely that without changes those implementations will pass these tests.

In my opinion that would be a very bad direction for the project. This will likely open the door to significant divergence between RMW impl. and leaving a lot of code untested which ultimately leads to reduced quality (just my 2 ct).

hidmic:
It has been decided that we will update Fast-RTPS, CycloneDDS and RTI Connext RMWs accordingly. No need for these restrictions anymore. Closing.

PR #119, opened 2020-08-20
Add publisher/subscription matched count API test coverage.

hidmic:
Connected to https://github.com/ros2/rmw/pull/262. Depends on https://github.com/ros2/rmw_fastrtps/pull/424, https://github.com/ros2/rmw_connext/pull/451, and https://github.com/ros2/rmw_cyclonedds/pull/223.

### DIFF
--- a/test_rmw_implementation/CMakeLists.txt
+++ b/test_rmw_implementation/CMakeLists.txt
@@ -34,6 +34,7 @@ if(BUILD_TESTING)
     )
     target_compile_definitions(test_init_shutdown${target_suffix}
       PUBLIC "RMW_IMPLEMENTATION=${rmw_implementation}")
+
     ament_target_dependencies(test_init_shutdown${target_suffix}
       osrf_testing_tools_cpp rcutils rmw rmw_implementation
     )
@@ -65,7 +66,7 @@ if(BUILD_TESTING)
     target_compile_definitions(test_publisher${target_suffix}
       PUBLIC "RMW_IMPLEMENTATION=${rmw_implementation}")
     ament_target_dependencies(test_publisher${target_suffix}
-      rcutils rmw rmw_implementation test_msgs
+      osrf_testing_tools_cpp rcutils rmw rmw_implementation test_msgs
     )
 
     ament_add_gtest(test_subscription${target_suffix}
@@ -75,7 +76,7 @@ if(BUILD_TESTING)
     target_compile_definitions(test_subscription${target_suffix}
       PUBLIC "RMW_IMPLEMENTATION=${rmw_implementation}")
     ament_target_dependencies(test_subscription${target_suffix}
-      rcutils rmw rmw_implementation test_msgs
+      osrf_testing_tools_cpp rcutils rmw rmw_implementation test_msgs
     )
 
     ament_add_gtest(test_serialize_deserialize${target_suffix}
--- a/test_rmw_implementation/CMakeLists.txt
+++ b/test_rmw_implementation/CMakeLists.txt
@@ -34,7 +34,6 @@ if(BUILD_TESTING)
     )
     target_compile_definitions(test_init_shutdown${target_suffix}
       PUBLIC "RMW_IMPLEMENTATION=${rmw_implementation}")
-
     ament_target_dependencies(test_init_shutdown${target_suffix}
       osrf_testing_tools_cpp rcutils rmw rmw_implementation
     )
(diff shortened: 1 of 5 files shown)


hidmic:
Matching pub/sub tests are not passing for `rmw_connext_cpp`. Arguably, these tests lack (1) a wait on the graph guard condition, and (2) an upper bound for repeated waits.

test_rmw_implementation/test/test_publisher.cpp
```diff
@@ -245,8 +252,123 @@ TEST_F(CLASSNAME(TestPublisherUse, RMW_IMPLEMENTATION), get_actual_qos) {
...
-  EXPECT_EQ(rmw_qos_profile_default.reliability, actual_qos_profile.reliability);
-  EXPECT_EQ(rmw_qos_profile_default.durability, actual_qos_profile.durability);
+  EXPECT_EQ(qos_profile.history, actual_qos_profile.history);
+  EXPECT_EQ(qos_profile.depth, actual_qos_profile.depth);
+  EXPECT_EQ(qos_profile.reliability, actual_qos_profile.reliability);
+  EXPECT_EQ(qos_profile.durability, actual_qos_profile.durability);
+}
+
+TEST_F(
+  CLASSNAME(TestPublisherUse, RMW_IMPLEMENTATION),
```
brawner, line 262: This alignment is strange since it aligns with the code. Is there not a way to adjust the indentation, or move the curly brace to a new line?

This happens elsewhere as well.

test_rmw_implementation/test/testing_macros.hpp
```diff
@@ -0,0 +1,31 @@
...
+#include <chrono>
+#include <thread>
+
+/// Retry until `timeout` expires, sleeping for `delay` in between attempts.
+/*
+ * \note Time is measured against OS provided steady clock.
+ */
+#define SLEEP_AND_RETRY_UNTIL(delay, timeout) for ( \
+    auto loop_start_time = std::chrono::steady_clock::now(), \
+    iteration_start_time = loop_start_time; \
```
brawner, line 27: Why is `iteration_start_time` needed? Why can't you just use `std::this_thread::sleep_for(delay)` instead of `sleep_until`?

test_rmw_implementation/CMakeLists.txt
```diff
@@ -34,6 +34,7 @@ if(BUILD_TESTING)
     )
     target_compile_definitions(test_init_shutdown${target_suffix}
       PUBLIC "RMW_IMPLEMENTATION=${rmw_implementation}")
+
```
brawner, line 37: This added line can be removed

test_rmw_implementation/test/testing_macros.hpp
hidmic: To avoid the time it takes to actually execute the loop code from adding up. Also, [`std::this_thread::sleep_for`](https://en.cppreference.com/w/cpp/thread/sleep_for) makes no guarantees as to which clock is used.

test_rmw_implementation/CMakeLists.txt
hidmic: Done in https://github.com/ros2/rmw_implementation/pull/119/commits/da381880ea605ff00ee45481e64b1759f2d57594.

test_rmw_implementation/test/test_publisher.cpp
hidmic: That's `uncrustify`... I ended up shortening the test case name in https://github.com/ros2/rmw_implementation/pull/119/commits/da381880ea605ff00ee45481e64b1759f2d57594.

test_rmw_implementation/test/testing_macros.hpp
brawner: I guess I mean something like:
```c++
#define SLEEP_AND_RETRY_UNTIL(delay, timeout) \
  for (auto start = std::chrono::steady_clock::now(); \ 
       std::chrono::steady_clock::now() - start < timeout; \
       std::this_thread::sleep_for(delay))
```

If you'd rather leave it as is, then I'm fine with it. It just seems like your use case doesn't require such precision.

hidmic: I didn't want to lose retries due to non-negligible execution times, but it is true that these timeouts are pretty arbitrary. 

Sure, why not. See https://github.com/ros2/rmw_implementation/pull/119/commits/4d85be9ec0d7244681f16beec420e5b2b0666127.

hidmic:
CI up to `test_rmw_implementation`, against all Tier 1 RMW implementations:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12070/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7095/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9844/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12001/)

hidmic:
@ros-pull-request-builder retest this please

hidmic:
@ros-pull-request-builder retest this please

hidmic:
Finally!

PR #127, opened 2020-09-16
Added benchmark test to rmw_implementation

ahcorde:
Most of the content here builds on rcpputils' shared library code and doesn't add any complexity worth testing. However, the preload operation lists quite a lot of symbols, and we should track the performance of that operation.

I tested `rmw_init()` which is in charge of fetch the initial symbols


ahcorde:
Compiling up-to `test_rmw_implementation` and testing `test_rmw_implementation`

With `-DAMENT_RUN_PERFORMANCE_TESTS=ON`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12322/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7291/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10035/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12224/)

cottsay (changes requested):
This is really good - I like that it runs for each of the RMWs. Would you mind renaming it to `rmw_init` since it's testing a lot more than just the prefetch function?

There is one big caveat here: library loading is only happening for the first call to `rmw_init`. For example, if one of the RMW libraries has a massive amount of shared libraries linked to it, we'll only pay the price of loading them all once, and then each iteration after that won't need to do it again. After looking at the implementation, I don't think we want to expose the additional API on `rmw_implementation` that would allow us to explicitly unload the library during each timing iteration, so these results will always be a best-case scenario.

Working off from this PR and also #135, I was able to confirm that forcing the library load causes the `rmw_init` duration to increase by around 20%, so I think that's something we'll want to track.

I'm really not sure why `rmw_implementation` has a separate package for tests, but if it's possible to move these tests into the `rmw_implementation` package itself, we can add some purpose-built functions to care of unloading.

ahcorde:
@cottsay, moved benchmark test to `rmw_implementation`

rmw_implementation/test/benchmark/benchmark_symbols.cpp
```diff
@@ -0,0 +1,37 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <memory>
+
+#include "performance_test_fixture/performance_test_fixture.hpp"
+
+#include "../../src/functions.cpp"
```
cottsay, line 19:
```suggestion
#include "../../src/functions.hpp"
```

We should be testing against the real compiler output instead of recompiling it ourself. This will require that you add `prefetch_symbols()` to `functions.hpp`.

rmw_implementation/test/benchmark/benchmark_symbols.cpp
```diff
@@ -0,0 +1,37 @@
...
+
+using performance_test_fixture::PerformanceTest;
+
+BENCHMARK_F(PerformanceTest, prefetch_symbols)(benchmark::State & st)
+{
+  for (auto _ : st) {
+    std::shared_ptr<rcpputils::SharedLibrary> lib = load_library();
+    prefetch_symbols();
```
cottsay, line 27:
```suggestion
    prefetch_symbols();
```

`load_library` doesn't cache the library as a global, so when `prefetch_symbols` runs, the library is loaded again by `get_library`. So this explicit load is extra.

cottsay (changes requested):

rmw_implementation/test/benchmark/benchmark_symbols.cpp
cottsay: In the interest of moving this PR along, I addressed this with 0507fc2.

rmw_implementation/test/benchmark/benchmark_symbols.cpp
cottsay: In the interest of moving this PR along, I addressed this with 0507fc2.

cottsay (approved):

PR #133, opened 2020-09-21
Add rmw_get_serialization_format() smoke test.

hidmic:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12378/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7347/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10090/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12290/)

### DIFF
--- a/test_rmw_implementation/test/test_serialize_deserialize.cpp
+++ b/test_rmw_implementation/test/test_serialize_deserialize.cpp
@@ -37,6 +37,12 @@ class CLASSNAME (TestSerializeDeserialize, RMW_IMPLEMENTATION) : public ::testin
 {
 };
 
+TEST_F(CLASSNAME(TestSerializeDeserialize, RMW_IMPLEMENTATION), get_serialization_format) {
+  const char * serialization_format = rmw_get_serialization_format();
+  EXPECT_NE(nullptr, serialization_format);
+  EXPECT_STREQ(serialization_format, rmw_get_serialization_format());
+}
+
 TEST_F(CLASSNAME(TestSerializeDeserialize, RMW_IMPLEMENTATION), serialize_with_bad_arguments) {
   const rosidl_message_type_support_t * ts{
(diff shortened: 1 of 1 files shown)


test_rmw_implementation/test/test_serialize_deserialize.cpp
```diff
@@ -37,6 +37,12 @@ class CLASSNAME (TestSerializeDeserialize, RMW_IMPLEMENTATION) : public ::testin
 {
 };
 
+TEST_F(CLASSNAME(TestSerializeDeserialize, RMW_IMPLEMENTATION), get_serialization_format) {
+  const char * serialization_format = rmw_get_serialization_format();
+  EXPECT_NE(nullptr, serialization_format);
+  EXPECT_STREQ(serialization_format, rmw_get_serialization_format());
```
brawner, line 43: You're checking `rmw_get_serialization_format()` equals itself?

hidmic: Checking that the outcome is not null, and that two subsequent calls yield the same outcome -- a poor man's _is this constant_ predicate. I don't feel strongly about it.

Lobotuerk (approved):
LGTM

hidmic:
@brawner if you don't mind, I'll move forward.

test_rmw_implementation/test/test_serialize_deserialize.cpp
brawner: I think if you're trying to ensure that the format string is identical, you should be asking the question why would it change? Would you be worried it is it based on some clock-time? Then it's unlikely to change in two lines of code. This test only seems to catch the case it would change due to some random internal state, which seems strange?

I think you should add a comment describing why this is necessary.

PR #143, opened 2020-09-28
Expose a semi-private function for unloading the RMW lib

cottsay:
This can be used to test the loading sequence for each RMW. I can already see that there is some investigation to do for Cyclone DDS.

~~I also renamed the `symbol_rmw_init` global variable to follow our naming convention and made it static.~~

~~https://index.ros.org/doc/ros2/Contributing/Code-Style-Language-Versions/#variable-naming~~

### DIFF
--- a/rmw_implementation/src/functions.cpp
+++ b/rmw_implementation/src/functions.cpp
@@ -42,7 +42,6 @@
 #define STRINGIFY(s) STRINGIFY_(s)
 
 static std::shared_ptr<rcpputils::SharedLibrary> g_rmw_lib = nullptr;
-static void * g_symbol_rmw_init = nullptr;
 
 std::shared_ptr<rcpputils::SharedLibrary>
 load_library()
@@ -127,13 +126,6 @@ void *
 get_symbol(const char * symbol_name)
 {
   return lookup_symbol(get_library(), symbol_name);
-}
-
-void
-unload_library()
-{
-  g_symbol_rmw_init = nullptr;
-  g_rmw_lib.reset();
 }
 
 #ifdef __cplusplus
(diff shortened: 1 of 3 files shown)


cottsay:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12505/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7473/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10217/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12425/)

rmw_implementation/src/functions.cpp
```diff
@@ -679,21 +688,19 @@ void prefetch_symbols(void)
...
-void * symbol_rmw_init = nullptr;
-
 rmw_ret_t
 rmw_init(const rmw_init_options_t * options, rmw_context_t * context)
 {
   prefetch_symbols();
-  if (!symbol_rmw_init) {
-    symbol_rmw_init = get_symbol("rmw_init");
+  if (!g_symbol_rmw_init) {
+    g_symbol_rmw_init = get_symbol("rmw_init");
```
hidmic, line 696: @cottsay meta: you may want to revisit `CALL_SYMBOL` and `RMW_INTERFACE_FN` macros as well for global variable naming updates.

rmw_implementation/src/functions.cpp
```diff
@@ -127,6 +129,13 @@ get_symbol(const char * symbol_name)
   return lookup_symbol(get_library(), symbol_name);
 }
 
+void
+unload_library()
+{
+  g_symbol_rmw_init = nullptr;
```
hidmic, line 135: @cottsay hmm what about all the other symbols?

rmw_implementation/test/benchmark/benchmark_symbols.cpp
```diff
@@ -23,8 +23,8 @@ using performance_test_fixture::PerformanceTest;
 BENCHMARK_F(PerformanceTest, prefetch_symbols)(benchmark::State & st)
 {
   for (auto _ : st) {
-    std::shared_ptr<rcpputils::SharedLibrary> lib = load_library();
     prefetch_symbols();
+    unload_library();
```
hidmic, line 27: @cottsay meta: are we benchmarking the OS dynamic linker?

hidmic (commented):
Overall LGTM, but for a few comments and questions.

rmw_implementation/src/functions.cpp
cottsay: I just noticed that like 10 mins ago. I'm going to revert this part of the change - maybe we can follow up sometime to clean up the whole pattern at once.

rmw_implementation/src/functions.cpp
cottsay: I'll add them. Good call.

rmw_implementation/test/benchmark/benchmark_symbols.cpp
cottsay: I think it's less about benchmarking the linker, and more about benchmarking how long it takes to load an RMW. The tests are currently show inconsistencies in this process that are hinting at some nondeterminism somewhere, and I'm also seeing that Cyclone DDS is taking 10 times as long as Fast-DDS to perform this prefetch operation - something we'll absolutely want to investigate.

It's less about the micro-scale testing of "how good is the OS dynamic linker performing", and more about determining how much this costs in total, and comparing between RMWs.

hidmic: > I'm also seeing that Cyclone DDS is taking 10 times as long as Fast-DDS to perform this prefetch operation - something we'll absolutely want to investigate.

Uh, really? That's both unexpected _and_ interesting. Could it be static storage initialization within `rmw_cyclonedds_cpp` or any of its dependencies?

hidmic (approved):
LGTM, though I agree we should revisit symbol naming and macro use.

rmw_implementation/test/benchmark/benchmark_symbols.cpp
cottsay: On my local machine, yeah. I mean, we're talking about 1/100 of a second or so, but it's still interesting. Connext is pretty bad, too. The numbers look better on the CI machine though.

Let's wait for the tests running regularly in the dedicated benchmark job before we launch any investigations. I just wanted to call it out the emphasize why I think this is a really great benchmark to track.

cottsay:
After nulling-out each of the symbols:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12512/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7480/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10224/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12432/)

Note that there are a bunch of warnings on Windows. I'm pretty sure they're caused by this: https://github.com/ros2/rmw_implementation/pull/127#discussion_r496222228

cottsay:
And, one last time. The Windows warnings should be fixed now.
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12513/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7481/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10225/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12433/)

PR #147, opened 2020-09-30
Update Quality Declaration to QL1.

hidmic:

### DIFF
--- a/rmw_implementation/QUALITY_DECLARATION.md
+++ b/rmw_implementation/QUALITY_DECLARATION.md
@@ -2,9 +2,9 @@ This document is a declaration of software quality for the `rmw_implementation` 
 
 # `rmw_implementation` Quality Declaration
 
-The package `rmw_implementation` claims to be in the **Quality Level 3** category.
+The package `rmw_implementation` claims to be in the **Quality Level 1** category.
 
-Below are the rationales, notes, and caveats for this claim, organized by each requirement listed in the [Package Requirements for Quality Level 3 in REP-2004](https://www.ros.org/reps/rep-2004.html).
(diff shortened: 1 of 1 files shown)


rmw_implementation/QUALITY_DECLARATION.md
```diff
@@ -2,9 +2,9 @@ This document is a declaration of software quality for the `rmw_implementation`
 
 # `rmw_implementation` Quality Declaration
 
-The package `rmw_implementation` claims to be in the **Quality Level 3** category.
+The package `rmw_implementation` claims to be in the **Quality Level 1** category.
```
chapulina, line 5: I believe this package is blocked from level 1 by its dependencies. It can only be level 2 right now, like `rcpputils`, `rcutils` and `rmw`. 

I see that `rmw_implementation_cmake` is also mentioned in the dependencies section, but is that a runtime dependency? We haven't been tracking it.

Lobotuerk (approved):
LGTM

rmw_implementation/QUALITY_DECLARATION.md
hidmic: All good points: 

- For `rcutils`, we're lacking performance tests (or a justification to not have them) to bump it to QL1. 
- For `rcpputils`, we're a bit under 90% coverage and we're lacking performance tests (or a justification to not have them) to bump it to QL1.
- For `rmw`, we're lacking performance tests (or a justification to not have them) to bump it to QL1.

This constraints apply to implementations as well.

About `rmw_implementation_cmake`, I think it's a build tool, build, and build export dependency, but not a runtime nor test dependency. I also don't see clearly what's missing there for QL1 besides proper justifications for missing tests (unless we decide we want to start testing CMake code, which I presume it's not the case).

hidmic: I'll relax to QL2 to get this going.

PR #149, opened 2020-09-30
Fix up C functions to never throw.

hidmic:
This will reduce coverage somewhat, as the conditions on which these exceptions are generated cannot be reproduced (yet).

CI up to `rmw_implementation`, `test_rmw_implementation`, and `rcl`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12685/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7648/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10400/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12628/)

### DIFF
--- a/rmw_implementation/src/functions.cpp
+++ b/rmw_implementation/src/functions.cpp
@@ -96,7 +96,7 @@ get_library()
 }
 
 void *
-lookup_symbol(std::shared_ptr<rcpputils::SharedLibrary> lib, const char * symbol_name)
+lookup_symbol(std::shared_ptr<rcpputils::SharedLibrary> lib, const std::string & symbol_name)
 {
   if (!lib) {
     if (!rmw_error_is_set()) {
@@ -110,22 +110,28 @@ lookup_symbol(std::shared_ptr<rcpputils::SharedLibrary> lib, const char * symbol
       std::string library_path = lib->get_library_path();
       RMW_SET_ERROR_MSG_WITH_FORMAT_STRING(
         "failed to resolve symbol '%s' in shared library '%s'",
-        symbol_name, library_path.c_str());
-    } catch (const std::runtime_error &) {
+        symbol_name.c_str(), library_path.c_str());
+    } catch (const std::exception & e) {
       RMW_SET_ERROR_MSG_WITH_FORMAT_STRING(
-        "failed to resolve symbol '%s' in shared library",
-        symbol_name);
+        "failed to resolve symbol '%s' in shared library due to %s",
+        symbol_name.c_str(), e.what());
     }
     return nullptr;
   }
-
   return lib->get_symbol(symbol_name);
 }
 
 void *
 get_symbol(const char * symbol_name)
 {
-  return lookup_symbol(get_library(), symbol_name);
+  try {
+    return lookup_symbol(get_library(), symbol_name);
+  } catch (const std::exception & e) {
+    RMW_SET_ERROR_MSG_WITH_FORMAT_STRING(
+      "failed to get symbol '%s' due to %s",
+      symbol_name, e.what());
+    return nullptr;
+  }
 }
 
 #ifdef __cplusplus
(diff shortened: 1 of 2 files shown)


Lobotuerk (approved):
LGTM

fujitatomoya (approved):

ahcorde:
CI up to `rmw_implementation`, `test_rmw_implementation`, and `rcl`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12704/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7665/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10418/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12650/)

ahcorde (approved):

clalancette (approved):
Seems reasonable to me, @hidmic should we merge?

iuhilnehc-ynos:
Catching the exception seems reasonable to me,
but I think to use `const string& symbol_name` instead of `const char * symbol_name` for `lookup_symbol` is not good, which causes an extra calling constructor of `std::string`.
I know `rcpputils` use 
```
  RCPPUTILS_PUBLIC
  void *
  get_symbol(const std::string & symbol_name);
```
that the parameter type is `std::string`, but I think it's better to ask `rcpputils` to provide an extra method, such as
```
  RCPPUTILS_PUBLIC
  void *
  get_symbol(const char * symbol_name);
```
, not to update the parameter type for `lookup_symbol` in `rmw_implementation`.

What do you think about that?

hidmic:
@iuhilnehc-ynos that's reasonable, but it doesn't preclude the need for this patch. Unless we make those methods `noexcept`, there are other reasons (besides a failing `std::string` construction) for them to throw. We have to handle them all.

hidmic:
Running CI again, this time computing test coverage:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12937/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7886/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10653/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12886/)

Also for reference, a Linux build w/o test coverage:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12938/)

iuhilnehc-ynos:
@hidmic 

> that's reasonable, but it doesn't preclude the need for this patch. Unless we make those methods `noexcept`, there are other reasons (besides a failing `std::string` construction) for them to throw. We have to handle them all.

I mean https://github.com/ros2/rmw_implementation/blob/75588ac591c4d11039d3765b3ccad652c08c84c6/rmw_implementation/src/functions.cpp#L127-L134 is reasonable, but I don't know why updating https://github.com/ros2/rmw_implementation/blob/75588ac591c4d11039d3765b3ccad652c08c84c6/rmw_implementation/src/functions.cpp#L99 that will slow down `GET_SYMBOL` in the https://github.com/ros2/rmw_implementation/blob/75588ac591c4d11039d3765b3ccad652c08c84c6/rmw_implementation/src/functions.cpp#L612.

draft sequence: prefetch_symbols -> GET_SYMBOL(rmw_get_implementation_identifier) -> symbol_rmw_get_implementation_identifier = get_symbol( "rmw_get_implementation_identifier" ); -> lookup_symbol(get_library(), `std::string`("rmw_get_implementation_identifier")).

Could you reconsider whether the `lookup_symbol` should be updated?
Because I found that you had agreed with https://github.com/ros2/rcpputils/pull/110.

hidmic:
Alright, from what I can observe in above's partial coverage computations, this patch does not appreciably degrade coverage. I'll go ahead and merge. Let's see what nightlies have to say.

PR #166, opened 2020-11-25
Update QD to QL 1

brawner:
Signed-off-by: Stephen Brawner <brawner@gmail.com>

### DIFF
--- a/rmw_implementation/QUALITY_DECLARATION.md
+++ b/rmw_implementation/QUALITY_DECLARATION.md
@@ -2,9 +2,9 @@ This document is a declaration of software quality for the `rmw_implementation` 
 
 # `rmw_implementation` Quality Declaration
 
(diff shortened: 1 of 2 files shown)


chapulina (approved):
:tada:

rmw_implementation/QUALITY_DECLARATION.md
```diff
@@ -2,9 +2,9 @@ This document is a declaration of software quality for the `rmw_implementation`
 
 # `rmw_implementation` Quality Declaration
 
-The package `rmw_implementation` claims to be in the **Quality Level 2** category.
+The package `rmw_implementation` claims to be in the **Quality Level 1** category when it is used with a **Quality Level 1** middleware.
```
chapulina, line 5: I just realized that this bit here was probably not needed, since this package doesn't rely on the middleware.

PR #171, opened 2021-01-27
Set the return value of rmw_ret_t before entering the loop.

clalancette:
clang static analysis pointed out that if we never did *any*
iterations of the loop in SLEEP_AND_RETRY_UNTIL, then we
would be comparing an uninitialized value of rmw_ret_t with
RMW_RET_OK.  Just set it to RMW_RET_ERROR in this unlikely
case which will cause us to fail the test.


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13519/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8422/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11245/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13586/)

fujitatomoya (approved):

PR #173, opened 2021-02-01
Set the value of is_available before entering the loop.

clalancette:
clang static analysis pointed out that if we never did *any*
iterations of the loop in SLEEP_AND_RETRY_UNTIL, then we
would be comparing an uninitialized value of is_available with
false.  Just set it to false in this unlikely case.


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13566/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8457/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11287/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13628/)

fujitatomoya (approved):

PR #174, opened 2021-02-03
Added tampered buffer deserialization tests

MiguelCompany:
Add tests for deserialization of a corrupted buffer.

The message type has been chosen on purpose to be `UnboundedSequences` in order to check the fixes on ros2/rmw_fastrtps#505 and eProsima/Fast-CDR#98. 

They perform the following sequence:
1. Serialization of a default constructed UnboundedSequences message, this would produce a buffer with a lot of zero bytes indicating the length of each sequence, which are 0.
2. Modify the resulting buffer to all FF's. This may produce crashes as they will be interpreted as sequences of `2^32 - 1` elements and either allocation errors or access beyond the end of the serialized buffer may occur
3. Expect the deserialization to fail

These tests fail with `rmw_fastrtps_cpp` and `rmw_fastrtps_dynamic_cpp` unless Fast CDR is updated to v1.0.19 and ros2/rmw_fastrtps#505 is applied

These new tests fail locally on `rmw_cyclonedds`, so it may need to be updated

### DIFF
--- a/test_rmw_implementation/test/test_serialize_deserialize.cpp
+++ b/test_rmw_implementation/test/test_serialize_deserialize.cpp
@@ -23,6 +23,8 @@
 
 #include "test_msgs/msg/basic_types.h"
 #include "test_msgs/msg/basic_types.hpp"
+#include "test_msgs/msg/unbounded_sequences.h"
+#include "test_msgs/msg/unbounded_sequences.hpp"
 
 #include "./allocator_testing_utils.h"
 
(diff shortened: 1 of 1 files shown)


MiguelCompany:
@eboasson FYI, it seems there is an issue on `rmw_cyclonedds` with the deserialization of corrupted buffers.

eboasson:
> @eboasson FYI, it seems there is an issue on `rmw_cyclonedds` with the deserialization of corrupted buffers.

Thanks, yes, the ROS2-specific deserialiser is a bit cavalier. The planned clean up still hasn't happened.

PR #175, opened 2021-02-03
Make sure to initialize the rmw_message_sequence after init.

clalancette:
The documentation for "rmw_take_sequence" says:

"Given `message_sequence` must be a valid message sequence, initialized
by rmw_message_sequence_init() and populated with ROS messages whose
type matches the message type support registered with the `subscription`
on creation."

As it stands, the take_sequence test fulfills the first part of that
(it was initialized by rmw_message_sequence_init()), but doesn't fulfill
the second part (populated by ROS messages).  In certain situations,
this could lead to an aborted test later on.  Make sure to do the
initialization to a known type before calling rmw_take_sequence.

This fixes the test part of the problem in https://github.com/ros2/rmw_cyclonedds/issues/279 .  There is still some follow-up work to do in the RMWs, but that can be done separately.

### DIFF
--- a/test_rmw_implementation/test/test_subscription.cpp
+++ b/test_rmw_implementation/test/test_subscription.cpp
@@ -24,6 +24,7 @@
 #include "rmw/error_handling.h"
 
 #include "test_msgs/msg/basic_types.h"
+#include "test_msgs/msg/strings.h"
 #include "./config.hpp"
 #include "./testing_macros.hpp"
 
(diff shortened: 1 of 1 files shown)


test_rmw_implementation/test/test_subscription.cpp
```diff
@@ -530,6 +531,12 @@ TEST_F(CLASSNAME(TestSubscriptionUse, RMW_IMPLEMENTATION), take_sequence) {
   rmw_message_sequence_t sequence = rmw_get_zero_initialized_message_sequence();
   rmw_ret_t ret = rmw_message_sequence_init(&sequence, count, &allocator);
   EXPECT_EQ(RMW_RET_OK, ret) << rmw_get_error_string().str;
+
+  auto seq = test_msgs__msg__Strings__Sequence__create(count);
```
cottsay, line 535: Shouldn't this have an accompanying `test_msgs__msg__Strings__Sequence__destroy(seq)`?

clalancette: Absolutely, great catch.  Should be fixed in 3db94ad .  Thanks for the review.

cottsay (approved):

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13625/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8507/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11340/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13685/)

clalancette:
CI is green, merging.  Thanks for the review.

PR #181, opened 2021-02-25
add content-filtered-topic interfaces

iuhilnehc-ynos:
Related to https://github.com/ros2/design/pull/282 to add content-filtered-topic interfaces.

### DIFF
--- a/test_rmw_implementation/test/test_subscription.cpp
+++ b/test_rmw_implementation/test/test_subscription.cpp
@@ -1254,11 +1254,10 @@ TEST_F(CLASSNAME(TestContentFilterSubscriptionUse, RMW_IMPLEMENTATION), get_cont
   if (sub->is_cft_enabled) {
     EXPECT_EQ(RMW_RET_OK, ret);
     EXPECT_STREQ(options.filter_expression, filter_expression);
-    ASSERT_NE(nullptr, options.expression_parameters);
-    ASSERT_EQ(expression_parameters_count, options.expression_parameters->size);
-    for (size_t i = 0; i < options.expression_parameters->size; ++i) {
+    ASSERT_EQ(expression_parameters_count, options.expression_parameters.size);
+    for (size_t i = 0; i < options.expression_parameters.size; ++i) {
       EXPECT_STREQ(
-        options.expression_parameters->data[i],
+        options.expression_parameters.data[i],
         expression_parameters[i]);
     }
     EXPECT_EQ(
--- a/test_rmw_implementation/test/test_subscription.cpp
+++ b/test_rmw_implementation/test/test_subscription.cpp
@@ -22,6 +22,7 @@
 
 #include "rmw/rmw.h"
 #include "rmw/error_handling.h"
+#include "rmw/subscription_content_filter_options.h"
 
 #include "test_msgs/msg/basic_types.h"
 #include "test_msgs/msg/strings.h"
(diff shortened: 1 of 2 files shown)


fujitatomoya (commented):
LGTM with test.

i think it would be nice to add test for both interfaces, along with `rmw_take`.

at this moment, no implementation supports those interfaces. but once [rmw_connextdds](https://github.com/rticommunity/rmw_connextdds) is merged, we can enable and adjust the test.

iuhilnehc-ynos:
> LGTM with test.
> 
> i think it would be nice to add test for both interfaces, along with `rmw_take`.
> 
> at this moment, no implementation supports those interfaces. but once [rmw_connextdds](https://github.com/rticommunity/rmw_connextdds) is merged, we can enable and adjust the test.

OK, I'll add this kind of test into `rcl` by using a flag, such as `is_connextdds` or `is_vendor_support_cft`.

iuhilnehc-ynos:
@fujitatomoya 

I have added some test cases in `rcl`, if you're free, you could review https://github.com/ros2/rcl/pull/894/commits/be5a640aee067ba6fec2b83c2fad781d38eef94e

fujitatomoya:
@audrow @mjcarroll 
CC: @ivanpauno @wjwwood 

Could you review this PR. (asking cz i see your names as maintainer.)

ivanpauno:
> Could you review this PR. (asking cz i see your names as maintainer.)

I'm planning to continue reviewing the PRs after I get concensus with the team about the plan mentioned in https://github.com/ros2/rmw/pull/302#issuecomment-1055642455.
I have already added the topic for discussion in a meeting next Tuesday.

test_rmw_implementation/test/test_subscription.cpp
```diff
@@ -870,6 +871,143 @@ TEST_F(
...
+  OSRF_TESTING_TOOLS_CPP_SCOPE_EXIT(
+  {
+    EXPECT_EQ(
+      RMW_RET_OK,
+      rmw_subscription_content_filter_options_fini(&options, &allocator));
+  });
+
+  ret = rmw_subscription_set_content_filter(sub, &options);
+  if (ret != RMW_RET_UNSUPPORTED) {
+    ASSERT_EQ(RMW_RET_OK, ret);
```
asorbini, line 972: The test should sleep for a short amount (e.g. 500ms) to allow for the changed filter to propagate from the reader to the writer, otherwise there is a chance for the next published samples to be filtered by the old filter.

test_rmw_implementation/test/test_subscription.cpp
```diff
@@ -1041,3 +1179,540 @@ bool operator==(const rmw_message_info_t & m1, const rmw_message_info_t & m2)
...
+    OSRF_TESTING_TOOLS_CPP_SCOPE_EXIT(
+    {
+      EXPECT_EQ(
+        RMW_RET_OK,
+        rmw_subscription_content_filter_options_fini(&options, &allocator));
+    });
+
+    EXPECT_EQ(
+      RMW_RET_OK,
+      rmw_subscription_set_content_filter(sub, &options));
```
asorbini, line 1392: Make sure to wait to allow for filter propagation.

test_rmw_implementation/test/test_subscription.cpp
```diff
@@ -1041,3 +1179,540 @@ bool operator==(const rmw_message_info_t & m1, const rmw_message_info_t & m2)
...
```
asorbini, line 1535: Make sure to wait to allow for filter propagation.

test_rmw_implementation/test/test_subscription.cpp
```diff
@@ -1041,3 +1179,540 @@ bool operator==(const rmw_message_info_t & m1, const rmw_message_info_t & m2)
...
```
asorbini, line 1679: Make sure to wait to allow for filter propagation.

asorbini (commented):
Test failures with `rmw_connextdds` should be resolved after adding a short sleep when changing the content filter.

test_rmw_implementation/test/test_subscription.cpp
iuhilnehc-ynos: I can add awaiting in the test, but I don't know if this API should be implemented as synchronous?
(NOTE: to wait in the test always causes a flaky error.)

test_rmw_implementation/test/test_subscription.cpp
iuhilnehc-ynos: I can add awaiting in the test, but I don't know if this API should be implemented as synchronous?
(NOTE: to wait in the test always causes a flaky error.)

iuhilnehc-ynos: It seems there is still a limitation in this API, maybe `connextDDS` could have to add the reader filter side.
I think users don't want to get old filter data after calling `rmw_subscription_set_content_filter` for a subscription.

fujitatomoya: I think that this is really important for user application.

this is said that a Writer side filter may be using outdated expression parameters.
if this is negative case, that should be okay because filter signature does not match by Reader side, the Reader will ignore the filter results and instead apply its local filter?
but if it is positive case, what would happen? since the message is already dropped by writer side with outdated filter expression?

iuhilnehc-ynos: > if this is negative case, that should be okay because filter signature does not match by Reader side, the Reader will ignore the filter results and instead apply its local filter?

Yes, I think so.

> if it is positive case, what would happen? since the message is already dropped by writer side with outdated filter expression?

IMO, users can accept that the message is dropped by writer side in this case.

fujitatomoya: > IMO, users can accept that the message is dropped by writer side in this case.

I just wondering that if i am developing application with CFT, i want to know when exactly the filtering expression and parameters are in effect even on DataWriter. instead, we have no idea outdated filtering still in effect... 

@asorbini @MiguelCompany any thoughts?

fujitatomoya (approved):

fujitatomoya:
CI is green, i will merge this into mainline.

- https://github.com/ros2/rmw/pull/302#issuecomment-1072005358
- https://github.com/ros2/rmw/pull/302#issuecomment-1072572088

PR #190, opened 2021-04-30
Remove unused headers

christophebedard:
https://github.com/ros2/rcutils/pull/340 deprecates `rcutils/get_env.h`. This PR doesn't need to wait until the `rcutils` PR is merged, though.

I simply removed the `#include` since nothing from that header was actually used.

### DIFF
--- a/rmw_implementation/test/test_functions.cpp
+++ b/rmw_implementation/test/test_functions.cpp
@@ -17,7 +17,6 @@
 #include <memory>
 
 #include "rcutils/env.h"
-#include "rcutils/get_env.h"
 #include "rcutils/testing/fault_injection.h"
 
 #include "rmw/error_handling.h"
--- a/rmw_implementation/src/functions.cpp
+++ b/rmw_implementation/src/functions.cpp
@@ -22,7 +22,6 @@
 
 #include "rcutils/allocator.h"
 #include "rcutils/format_string.h"
-#include "rcutils/get_env.h"
 #include "rcutils/types/string_array.h"
 
 #include "rcpputils/get_env.hpp"


rmw_implementation/test/test_functions.cpp
```diff
@@ -17,7 +17,6 @@
 #include <memory>
 
 #include "rcutils/env.h"
-#include "rcutils/get_env.h"
```
clalancette, line 20: Unfortunately, we can't do this here because `rcutils_get_env` is used below (at least, we can't do this prior to https://github.com/ros2/rcutils/pull/340 being merged).  So either this has to remain part of the larger series to move `rcutils_get_env` into env.h, or we have to switch the test code to use `rcpputils::get_env`.  Either is fine with me.

christophebedard: `rcutils/env.h` includes `rcutils/get_env.h`, though: https://github.com/ros2/rcutils/blob/ca42bc9f588e11ad40f5879dcb9a364b435d2cba/include/rcutils/env.h#L31

fujitatomoya (approved):

rmw_implementation/test/test_functions.cpp
clalancette: > `rcutils/env.h` includes `rcutils/get_env.h`, though: https://github.com/ros2/rcutils/blob/ca42bc9f588e11ad40f5879dcb9a364b435d2cba/include/rcutils/env.h#L31

Ah!  I totally missed that.  Thanks for the heads up.  I'll approve and run CI on this, then.

clalancette (approved):

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14541/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9297/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12206/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14647/)

PR #191, opened 2021-05-04
Wait for server in test_rmw_implementation service tests

hidmic:
Current tests assume service server and client in the same process discover each other immediately. That doesn't seem to be true for `rmw_connextdds` and thus [the flakes in amd64 Linux repeated CI](https://ci.ros2.org/view/nightly/job/nightly_linux_repeated/lastCompletedBuild/testReport/(root)/projectroot/test_service__rmw_connextdds/), which can be reproduced locally (FYI @asorbini). 

This patch amends this. 

Repeated CI up to `test_rmw_implementation`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14554/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9311/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12222/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14681/)

### DIFF
--- a/test_rmw_implementation/test/test_client.cpp
+++ b/test_rmw_implementation/test/test_client.cpp
@@ -386,25 +386,25 @@ TEST_F(CLASSNAME(TestClientUse, RMW_IMPLEMENTATION), service_server_is_available
 {
   bool is_available = false;
   rmw_ret_t ret = rmw_service_server_is_available(node, client, &is_available);
-  EXPECT_EQ(RMW_RET_OK, ret) << rmw_get_error_string().str;
+  ASSERT_EQ(RMW_RET_OK, ret) << rmw_get_error_string().str;
   ASSERT_FALSE(is_available);
-  rmw_reset_error();
 
   rmw_service_t * service = rmw_create_service(node, ts, service_name, &qos_profile);
-  ASSERT_NE(nullptr, client) << rcutils_get_error_string().str;
+  ASSERT_NE(nullptr, client) << rmw_get_error_string().str;
+  OSRF_TESTING_TOOLS_CPP_SCOPE_EXIT(
+  {
+    rmw_ret_t ret = rmw_destroy_service(node, service);
+    EXPECT_EQ(RMW_RET_OK, ret) << rmw_get_error_string().str;
+    rmw_reset_error();
+  });
 
   is_available = false;
   SLEEP_AND_RETRY_UNTIL(rmw_intraprocess_discovery_delay, rmw_intraprocess_discovery_delay * 10) {
     ret = rmw_service_server_is_available(node, client, &is_available);
-    EXPECT_EQ(RMW_RET_OK, ret) << rmw_get_error_string().str;
-    rmw_reset_error();
+    ASSERT_EQ(RMW_RET_OK, ret) << rmw_get_error_string().str;
     if (is_available) {
       break;
     }
   }
-  EXPECT_TRUE(is_available) << rmw_get_error_string().str;
-  rmw_reset_error();
-
-  ret = rmw_destroy_service(node, service);
-  EXPECT_EQ(RMW_RET_OK, ret) << rmw_get_error_string().str;
+  ASSERT_TRUE(is_available);
 }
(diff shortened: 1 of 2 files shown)


test_rmw_implementation/test/test_client.cpp
```diff
@@ -385,26 +385,23 @@ TEST_F(CLASSNAME(TestClientUse, RMW_IMPLEMENTATION), service_server_is_available
...
-    ret = rmw_service_server_is_available(node, client, &is_available);
-    if (RMW_RET_OK == ret && is_available) {
-      break;
-    }
-  }
-  EXPECT_EQ(ret, RMW_RET_OK) << rmw_get_error_string().str;
-  EXPECT_FALSE(is_available) << rmw_get_error_string().str;
+  rmw_ret_t ret = rmw_service_server_is_available(node, client, &is_available);
+  EXPECT_EQ(RMW_RET_OK, ret) << rmw_get_error_string().str;
+  ASSERT_FALSE(is_available);
```
clalancette, line 390: Just so I understand properly; we don't need to wait here because we never created a server at all, right?

test_rmw_implementation/test/test_client.cpp
```diff
@@ -385,26 +385,23 @@ TEST_F(CLASSNAME(TestClientUse, RMW_IMPLEMENTATION), service_server_is_available
...
   rmw_service_t * service = rmw_create_service(node, ts, service_name, &qos_profile);
   ASSERT_NE(nullptr, client) << rcutils_get_error_string().str;
+
+  is_available = false;
   SLEEP_AND_RETRY_UNTIL(rmw_intraprocess_discovery_delay, rmw_intraprocess_discovery_delay * 10) {
     ret = rmw_service_server_is_available(node, client, &is_available);
-    if (RMW_RET_OK == ret && is_available) {
+    EXPECT_EQ(RMW_RET_OK, ret) << rmw_get_error_string().str;
+    rmw_reset_error();
+    if (is_available) {
```
clalancette, line 401: I'm not sure that this is the right logic here (I also don't think the previous logic is correct either).

What we are doing here is looping for a specified amount of time, waiting to see if the service server has become available.  From what I understand from the documentation, `rmw_service_server_is_available` should *always* return `RMW_RET_OK`, unless the typesupport didn't match or an unexpected error occurred.  But in those latter 2 cases, that is clearly a failure of this test, and so further waiting doesn't seem warranted.  Thus, I would think this logic would be:

```
if (ret != RMW_RET_OK) {
    break;
}
if (is_available) {
    break;
}
```

And then we keep both expectations below.  What do you think?

(this same comment applies to both loops below)

test_rmw_implementation/test/test_client.cpp
hidmic: That's my recollection, yes.

test_rmw_implementation/test/test_client.cpp
hidmic: I'd be fine with it. We could also just `ASSERT_EQ(RMW_RET_OK, ret) << rmw_get_error_string().str;` right after calling `rmw_service_is_available`. Which one would you prefer?

clalancette: > I'd be fine with it. We could also just `ASSERT_EQ(RMW_RET_OK, ret) << rmw_get_error_string().str;` right after calling `rmw_service_is_available`. Which one would you prefer?

The downside to doing an `ASSERT` is that we'd never hit the cleanup.  While that doesn't matter *too* much in a test, it has the side-effect of making clang static analysis complain, which I'd rather avoid.

hidmic: Oh, but that's because the test is not doing clean-up properly ! See 2da26ad311836667e68f98f4ab5c32d42876882a.

hidmic:
Regular CI up to `test_rmw_implementation`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14571/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9327/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12234/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14696/)

clalancette (approved):

PR #198, opened 2021-11-02
Fix renamed `rcpputils` header

aprotyas:
`rcpputils/get_env.hpp` is renamed to `rcpputils/env.hpp` as part of https://github.com/ros2/rcpputils/pull/150, so the corresponding includes have been fixed in this PR.

Signed-off-by: Abrar Rahman Protyasha <aprotyas@u.rochester.edu>

### DIFF
--- a/rmw_implementation/src/functions.cpp
+++ b/rmw_implementation/src/functions.cpp
@@ -26,7 +26,7 @@
 #include "rcutils/format_string.h"
 #include "rcutils/types/string_array.h"
 
-#include "rcpputils/get_env.hpp"
+#include "rcpputils/env.hpp"
 #include "rcpputils/shared_library.hpp"
 
 #include "rmw/error_handling.h"


aprotyas:
The Rpr job fails as expected because the renamed header is not in a release yet.

CI from https://github.com/ros2/rcpputils/pull/150:

Repos file: https://gist.github.com/aprotyas/96874af9175859555af0144398e0bf25
Build/Test args: `--packages-above-and-dependencies rcpputils`
ROS distro: Rolling
Job: https://ci.ros2.org/job/ci_launcher/9263

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15548/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10232/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/13215/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15760/)

rmw_implementation/src/functions.cpp
```diff
@@ -26,7 +26,7 @@
 #include "rcutils/format_string.h"
 #include "rcutils/types/string_array.h"
 
-#include "rcpputils/get_env.hpp"
+#include "rcpputils/env.hpp"
```
clalancette, line 29: This isn't right, as far as I can tell.  We deprecated `rcutils/get_env.h`, not `rcpputils/get_env.hpp`.  So I don't think this is going to work.

clalancette (changes requested):

rmw_implementation/src/functions.cpp
clalancette: Oh, never mind me.  I didn't read the whole description, the whole point of this is to add that :).

clalancette (approved):

aprotyas:
Two approvals and CI is green, so I'll merge this in. Thanks for the reviews!

PR #202, opened 2021-12-20
Fix build when only one rmw implementation is available

ivanpauno:
After https://github.com/ros2/rmw_cyclonedds/pull/357, fix with only rmw_cyclonedds available were broken (I usually run CI with only one rmw impl to save time, except when it makes sense to test them all).
Example: https://github.com/ros2/rcl/pull/956#issuecomment-997050045.

These removes the old trick done in the configs extra file, and it instead reexports and interface library.
I could have kept the config extras file and add an IMPORTED target when appropriate there, but the end result I think is the same and the version is simpler.

Note:

I will run "one rmw" jobs for each rmw implementation to make sure this works in all cases, and also a multiple rmw job.


ivanpauno:
> I will run "one rmw" jobs for each rmw implementation to make sure this works in all cases, and also a multiple rmw job.

That seem to have worked locally, so running all the jobs.

---

Fastrtps only:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15801/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10479/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/13463/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16077/)

---

cyclonedds only:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15802/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10480/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/13464/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16078/)
Triggering a new build of ci_linux

---

connextdds only:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15803/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10481/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/13465/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16079/)

---

fastrtps/cyclonedds/connextdds combined:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15804/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10482/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/13466/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16080/)

ivanpauno:
I don't understand how this was passing locally, I guess I didn't clean up my build and install folders well enough ...
It seems that fastrtps and connexts are not exporting cmake targes, so I will first update that.

sloretz:
I think this and #201 are solving the same issue. Unfortunately, I'm having strange import issues in launch_xml with the other PR that I haven't understood yet :-/

ivanpauno:
> I think this and #201 are solving the same issue. Unfortunately, I'm having strange import issues in launch_xml with the other PR that I haven't understood yet :-/

Yes, it's the same issue.
I will close this one, as your PR avoids the issue I was running into altogether (https://github.com/ros2/rmw_implementation/pull/202#issuecomment-998153884).

If we modify rmw implementations to export targets correctly, the approach of this PR requires a bit less boilerplate, but that doesn't matter much.

PR #206, opened 2022-08-15
Add rmw_get_gid_for_client & tests

ihasdapie:
See: https://github.com/ros2/rmw/pull/327

Signed-off-by: Brian Chen <brian.chen@openrobotics.org>

### DIFF
--- a/test_rmw_implementation/test/test_unique_identifiers.cpp
+++ b/test_rmw_implementation/test/test_unique_identifiers.cpp
@@ -59,13 +59,16 @@ protected:
     pub = rmw_create_publisher(node, ts, topic_name, &qos_profile, &options);
     ASSERT_NE(nullptr, pub) << rmw_get_error_string().str;
 
-    constexpr char service_name[] = "test_service";
-    cli = rmw_create_client(node, srv_ts, service_name, &qos_profile);
+    constexpr char service_name[] = "/test_service0";
+    client = rmw_create_client(node, srv_ts, service_name, &qos_profile);
+    ASSERT_NE(nullptr, client) << rmw_get_error_string().str;
   }
 
   void TearDown() override
   {
     rmw_ret_t ret = rmw_destroy_publisher(node, pub);
+    EXPECT_EQ(RMW_RET_OK, ret) << rmw_get_error_string().str;
+    ret = rmw_destroy_client(node, client);
     EXPECT_EQ(RMW_RET_OK, ret) << rmw_get_error_string().str;
     ret = rmw_destroy_node(node);
     EXPECT_EQ(RMW_RET_OK, ret) << rmw_get_error_string().str;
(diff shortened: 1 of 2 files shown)


test_rmw_implementation/test/test_unique_identifiers.cpp
```diff
@@ -113,6 +119,41 @@ TEST_F(CLASSNAME(TestUniqueIdentifierAPI, RMW_IMPLEMENTATION), get_gid_with_bad_
...
+  const char * implementation_identifier = cli->implementation_identifier;
+  cli->implementation_identifier = "not-an-rmw-implementation-identifier";
+  ret = rmw_get_gid_for_client(cli, &gid);
+  cli->implementation_identifier = implementation_identifier;
+  EXPECT_EQ(RMW_RET_INCORRECT_RMW_IMPLEMENTATION, ret);
+  rmw_reset_error();
+  gids_are_equal = false;
+  ret = rmw_compare_gids_equal(&expected_gid, &gid, &gids_are_equal);
+  EXPECT_EQ(RMW_RET_OK, ret) << rmw_get_error_string().str;
+  EXPECT_TRUE(gids_are_equal);
```
ivanpauno, line 154: why?
`gid` should never have changed if the function called change, so it should still be equal to the original `expected_gid`

test_rmw_implementation/test/test_unique_identifiers.cpp
```diff
@@ -221,3 +262,52 @@ TEST_F(CLASSNAME(TestUniqueIdentifiersForMultiplePublishers, RMW_IMPLEMENTATION)
...
+  ASSERT_EQ(RMW_RET_OK, ret) << rmw_get_error_string().str;
+
+  ret = rmw_get_gid_for_client(cli2, &cli2_gid);
+  ASSERT_EQ(RMW_RET_OK, ret) << rmw_get_error_string().str;
+
+  bool are_equal = true;
+  ret = rmw_compare_gids_equal(&cli1_gid, &cli2_gid, &are_equal);
+  EXPECT_EQ(RMW_RET_OK, ret) << rmw_get_error_string().str;
+  EXPECT_FALSE(are_equal);
+}
```
ivanpauno, line 313: I'm not sure this is something we should we testing here, but it's okay to add it

test_rmw_implementation/test/test_unique_identifiers.cpp
jacobperron: That's correct. It is still equal to the original, hence we expect a true value. This test is checking that `gid` is not modified whenever we make a bad call to the getter. I think this logic is basically copied over from the tests for the publisher getter.

test_rmw_implementation/test/test_unique_identifiers.cpp
jacobperron: Yeah, it seems a bit odd to me too. I believe this was added to copy the test for the publisher gid in this file.

jacobperron:
I forgot to push my local changes, which address some bugs. @ivanpauno please take another look.

test_rmw_implementation/test/test_unique_identifiers.cpp
ivanpauno: ah sorry, I misread the code

ivanpauno (approved):
LGTM!

test_rmw_implementation/test/test_unique_identifiers.cpp
ihasdapie: Yeah, these methods (and the comment from the previous review) were just copied over to be symmetrical with the publisher gid tests

PR #208, opened 2022-10-08
Add rmw_count_clients,services & test

leeminju531:
this PR is for service counts using it's name like topic.
Related to https://github.com/ros2/ros2cli/pull/771

Signed-off-by: leeminju531 <dlalswn531@naver.com>

### DIFF
--- a/test_rmw_implementation/test/test_graph_api.cpp
+++ b/test_rmw_implementation/test/test_graph_api.cpp
@@ -946,7 +946,6 @@ TEST_F(CLASSNAME(TestGraphAPI, RMW_IMPLEMENTATION), count_clients_with_bad_argum
 TEST_F(CLASSNAME(TestGraphAPI, RMW_IMPLEMENTATION), count_services_with_bad_arguments) {
   size_t count = 0u;
   constexpr char service_name[] = "/test_service";
-  
   // A null node is an invalid argument.
   EXPECT_EQ(RMW_RET_INVALID_ARGUMENT, rmw_count_services(nullptr, service_name, &count));
   rmw_reset_error();
--- a/test_rmw_implementation/test/test_graph_api.cpp
+++ b/test_rmw_implementation/test/test_graph_api.cpp
@@ -913,3 +913,61 @@ TEST_F(CLASSNAME(TestGraphAPI, RMW_IMPLEMENTATION), count_subscribers_with_bad_a
   EXPECT_EQ(RMW_RET_INVALID_ARGUMENT, rmw_count_subscribers(node, topic_name, nullptr));
   rmw_reset_error();
 }
+
+TEST_F(CLASSNAME(TestGraphAPI, RMW_IMPLEMENTATION), count_clients_with_bad_arguments) {
+  size_t count = 0u;
+  constexpr char service_name[] = "/test_service";
+
+  // A null node is an invalid argument.
+  EXPECT_EQ(RMW_RET_INVALID_ARGUMENT, rmw_count_clients(nullptr, service_name, &count));
+  rmw_reset_error();
+
+  // A node from a different implementation cannot be used to query.
+  const char * implementation_identifier = node->implementation_identifier;
+  node->implementation_identifier = "not-an-rmw-implementation-identifier";
+  EXPECT_EQ(RMW_RET_INCORRECT_RMW_IMPLEMENTATION, rmw_count_clients(node, service_name, &count));
+  node->implementation_identifier = implementation_identifier;
+  rmw_reset_error();
+
+  // A null topic name is an invalid argument.
+  EXPECT_EQ(RMW_RET_INVALID_ARGUMENT, rmw_count_clients(node, nullptr, &count));
+  rmw_reset_error();
+
+  // An invalid topic name is an invalid argument.
+  constexpr char invalid_service_name[] = "not a valid topic name !";
+  EXPECT_EQ(RMW_RET_INVALID_ARGUMENT, rmw_count_clients(node, invalid_service_name, &count));
+  rmw_reset_error();
+
+  // A null count is an invalid argument.
+  EXPECT_EQ(RMW_RET_INVALID_ARGUMENT, rmw_count_clients(node, service_name, nullptr));
+  rmw_reset_error();
+}
+
+TEST_F(CLASSNAME(TestGraphAPI, RMW_IMPLEMENTATION), count_services_with_bad_arguments) {
+  size_t count = 0u;
+  constexpr char service_name[] = "/test_service";
+  
+  // A null node is an invalid argument.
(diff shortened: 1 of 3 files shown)


fujitatomoya (approved):
lgtm with consensus on https://github.com/ros2/rmw/pull/334#discussion_r1057986256

methylDragon:
rmw interface changes have been merged
- https://github.com/ros2/rmw/pull/334

test_rmw_implementation/test/test_graph_api.cpp
```diff
@@ -913,3 +916,103 @@ TEST_F(CLASSNAME(TestGraphAPI, RMW_IMPLEMENTATION), count_subscribers_with_bad_a
...
+TEST_F(CLASSNAME(TestGraphAPI, RMW_IMPLEMENTATION), count_clients_and_services) {
+  constexpr char service_name[] = "/test_service";
+  const rosidl_service_type_support_t * ts =
+    ROSIDL_GET_SRV_TYPE_SUPPORT(test_msgs, srv, BasicTypes);
+  size_t count = 0u;
+
+  rmw_service_t * srv = rmw_create_service(
+    other_node, ts, service_name, &rmw_qos_profile_services_default);
+  ASSERT_NE(nullptr, srv) << rmw_get_error_string().str;
+  std::this_thread::sleep_for(std::chrono::milliseconds(100));
```
clalancette, line 986: I'd really prefer not to have hard-coded sleep times in the tests.  Over time we've found that they are both too long and too short.

They are too long because if the action completes very quickly, we are wasting time.  They are too short because of timings within individual machines, and in CI, which leads to them being flaky.

Elsewhere, what we do is a tight loop where we check for the condition.  If the condition is satisfied, we break out of the loop.  If it is not satisfied, we sleep for a very short amount of time (1 or 10 milliseconds), then check again.  And then we have an overall timeout (2 or 5 seconds), just to ensure that we don't run forever in the case of a bug.  I'd much prefer if we did something closer to that here.

clalancette (changes requested):

test_rmw_implementation/test/test_graph_api.cpp
leeminju531: I really appreciate your cool explanation! fixed

test_rmw_implementation/test/test_graph_api.cpp
```diff
@@ -913,3 +915,151 @@ TEST_F(CLASSNAME(TestGraphAPI, RMW_IMPLEMENTATION), count_subscribers_with_bad_a
...
+  node->implementation_identifier = "not-an-rmw-implementation-identifier";
+  EXPECT_EQ(RMW_RET_INCORRECT_RMW_IMPLEMENTATION, rmw_count_clients(node, service_name, &count));
+  node->implementation_identifier = implementation_identifier;
+  rmw_reset_error();
+
+  // A null topic name is an invalid argument.
+  EXPECT_EQ(RMW_RET_INVALID_ARGUMENT, rmw_count_clients(node, nullptr, &count));
+  rmw_reset_error();
+
+  // An invalid topic name is an invalid argument.
```
clalancette, line 938:
```suggestion
  // An invalid service name is an invalid argument.
```

test_rmw_implementation/test/test_graph_api.cpp
```diff
@@ -913,3 +915,151 @@ TEST_F(CLASSNAME(TestGraphAPI, RMW_IMPLEMENTATION), count_subscribers_with_bad_a
...
+  rmw_reset_error();
+
+  // A node from a different implementation cannot be used to query.
+  const char * implementation_identifier = node->implementation_identifier;
+  node->implementation_identifier = "not-an-rmw-implementation-identifier";
+  EXPECT_EQ(RMW_RET_INCORRECT_RMW_IMPLEMENTATION, rmw_count_services(node, service_name, &count));
+  node->implementation_identifier = implementation_identifier;
+  rmw_reset_error();
+
+  // A null topic name is an invalid argument.
```
clalancette, line 962:
```suggestion
  // A null service name is an invalid argument.
```

test_rmw_implementation/test/test_graph_api.cpp
```diff
@@ -913,3 +915,151 @@ TEST_F(CLASSNAME(TestGraphAPI, RMW_IMPLEMENTATION), count_subscribers_with_bad_a
...
+    ROSIDL_GET_SRV_TYPE_SUPPORT(test_msgs, srv, BasicTypes);
+  size_t count = 0u;
+  rmw_ret_t ret;
+  rmw_service_t * srv = rmw_create_service(
+    other_node, ts, service_name, &rmw_qos_profile_services_default);
+  ASSERT_NE(nullptr, srv) << rmw_get_error_string().str;
+
+  SLEEP_AND_RETRY_UNTIL(rmw_intraprocess_discovery_delay, rmw_intraprocess_discovery_delay * 10) {
+    ret = rmw_count_services(node, service_name, &count);
+    ASSERT_EQ(RMW_RET_OK, ret) << rmw_get_error_string().str;
```
clalancette, line 988: This won't work, because if we fail once, that will cause everything after this to fail as well.  What we should do is remove this line from the loop, and then after the loop have an ASSERT that looks something like:

```
ASSERT_EQ(1u == count);
```

The same goes for other uses of `SLEEP_AND_RETRY_UNTIL` below.

clalancette (changes requested):

test_rmw_implementation/test/test_graph_api.cpp
leeminju531: This tight loop checks both whether the return code is successful and whether the intended count is met. Despite this loop being used to wait for the graph update, the `rmw_count_services` should still return a successful return code even if the intended count is not reached. Therefore, it seems appropriate that a failure occurring once in this line could result in the test failing. What do you think @clalancette ?

leeminju531:
@clalancette could you check https://github.com/ros2/rmw_implementation/pull/208#discussion_r1327984686 when you have a chance.

test_rmw_implementation/test/test_graph_api.cpp
clalancette: Oh yes, you are correct.  Thank you for pointing that out.

clalancette (approved):

clalancette:
This looks good now, thanks for the patience @leeminju531 .  I went ahead and rebased all of the PRs here onto the latest, and now I'm going to run CI with the following PRs only:

* https://github.com/ros2/rmw_implementation/pull/208

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19582/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14093/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20307/)

PR #214, opened 2023-01-27
Update rmw_implementation to C++17.

clalancette:
The two reasons to do this are:

1. So that we can compile rmw_implementation with the clang static analyzer. As of clang++-14 (what is in Ubuntu 22.04), the default still seems to be C++14, so we need to specify C++17 so that new things in the rclcpp headers work properly.
2. So we can build with a newer version of gtest which requires this.

Further, due to reasons I don't fully understand, I needed to set CMAKE_CXX_STANDARD_REQUIRED in order for clang to really use that version. So set this as well.

### DIFF
--- a/rmw_implementation/CMakeLists.txt
+++ b/rmw_implementation/CMakeLists.txt
@@ -2,9 +2,10 @@ cmake_minimum_required(VERSION 3.5)
 
 project(rmw_implementation)
 
-# Default to C++14
+# Default to C++17
 if(NOT CMAKE_CXX_STANDARD)
-  set(CMAKE_CXX_STANDARD 14)
+  set(CMAKE_CXX_STANDARD 17)
+  set(CMAKE_CXX_STANDARD_REQUIRED ON)
 endif()
 
 if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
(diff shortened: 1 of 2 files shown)


fujitatomoya (approved):
quick question, are you going to apply the same change to other projects for clang static analyzer? why there 2 projects are picked up?

clalancette:
> quick question, are you going to apply the same change to other projects for clang static analyzer?

Yeah, I have a whole series of these queued up.  See for example https://github.com/ros2/rosidl_typesupport/pull/131 , https://github.com/ros2/rosidl_typesupport_fastrtps/pull/94 , etc (and there are more that I haven't even opened yet).

> why there 2 projects are picked up?

Sorry, I don't quite understand this question.  Can you rephrase?

fujitatomoya:
> Sorry, I don't quite understand this question. Can you rephrase?

I just saw the two PRs, so that I came up with that question.

> Yeah, I have a whole series of these queued up. See for example https://github.com/ros2/rosidl_typesupport/pull/131 , https://github.com/ros2/rosidl_typesupport_fastrtps/pull/94 , etc (and there are more that I haven't even opened yet).

this answers my question too! thanks.

clalancette:
> this answers my question too! thanks.

You are welcome!

CI for this is in https://github.com/ros2/rosidl_typesupport/pull/131#issuecomment-1406672534, and since that is green and this is approved, I'm going to merge it in.

PR #223, opened 2024-01-31
Uncrustify update for noble.

marcoag:
Updating files to pass uncrustify test on noble.

### DIFF
--- a/test_rmw_implementation/test/test_client.cpp
+++ b/test_rmw_implementation/test/test_client.cpp
@@ -35,7 +35,7 @@
 # define CLASSNAME(NAME, SUFFIX) NAME
 #endif
 
-class CLASSNAME (TestClient, RMW_IMPLEMENTATION) : public ::testing::Test
+class CLASSNAME(TestClient, RMW_IMPLEMENTATION) : public ::testing::Test
 {
 protected:
   void SetUp() override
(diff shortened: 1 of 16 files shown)


clalancette:
Arg.  So it finally happened that Ubuntu 24.04 updated to the newer version of uncrustify.  I was hoping that wasn't going to happen, but here we are.

What's happening here is that the Rpr job is using the version of uncrustify that is in Ubuntu 22.04, which is uncrustify 0.72.  But it looks like Ubuntu 24.04 has now upgraded to uncrustify 0.78.1.  And those two versions of uncrustify don't seem to agree on the format.

As far as the uncrustify upgrade is concerned, I think that there are 2 goals:
1.  Pass the linter.
2.  Make it so that we can pass the linter on both Ubuntu 22.04 and 24.04.

That second goal is a bit tricky.  In the past, we've sometimes been able to find styles that satisfy both the old and new versions of uncrustify.  If we can't do that, then I think the other option is to update https://github.com/ament/uncrustify_vendor/blob/rolling/CMakeLists.txt so that it requires uncrustify 0.78.  That means that on Ubuntu 22.04, we'll start building it from source, but on Ubuntu 24.04 we'll just use the system version.  That should also fix the build on Windows and RHEL as well.

@marcoag What I'm going to suggest here is that you open a new issue on https://github.com/ros2/ros2 to track the upgrade to uncrustify.  From what I can tell, this is going to take several dozen PRs to fix, so we may as well have a central tracking place for them.

clalancette:
This is now obsoleted by #224, so closing it out.

PR #224, opened 2024-02-26
Compile the test_rmw_implementation tests fewer times.

clalancette:
We really only need to compile them once, and then we can run them with different RMW implementations.

This also neatly sidesteps an incompatibility between current uncrustify and the newer one.


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20361/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14769/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21112/)

mjcarroll (approved):

PR #229, opened 2024-04-24
Add mechanism to disable workaround for dependency groups

cottsay:
See ros-infrastructure/catkin_pkg#369

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20943/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15234/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21617/)

### DIFF
--- a/rmw_implementation/package.xml
+++ b/rmw_implementation/package.xml
@@ -21,15 +21,12 @@
   <depend>rcpputils</depend>
   <depend>rcutils</depend>
   <build_depend>rmw</build_depend>
-  <!--
-  Bloom does not support group_depend so entries below duplicate the group rmw_implementation_packages.
-  This ensures that binary packages have support for all of these rmw impl. enabled.
-  -->
-  <build_depend>rmw_connextdds</build_depend>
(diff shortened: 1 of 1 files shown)


nuclearsandwich (approved):

emersonknapp:
Is this all right to backport to Jazzy & Humble?
Here is me trying it, but I'm not sure if there are some considerations i might be missing

@mergifyio backport jazzy humble

cottsay:
There isn't a particularly great reason to backport this. We still aren't fully utilizing it in any of our tooling, so this has only been used for experiments thus far. That said, I don't see any harm in backporting it either.

emersonknapp:
In all my CIs over the last few years I've had to add really explicit rosdep skip-keys for all the variations on Connext, rmw_connext, etc. But if instead I have this in the package.xml, I can set the environment variable and am good to go. Largely because we are building rmw_implementation from source on a fork https://github.com/robograph-project/rmw_implementation that I'm trying to keep as minimally different as possible.

PR #266, opened 2025-10-13
RMW implementation library wrapper/shim 

emersonknapp:
## Description

(Note: I've hesitated on opening this contribution because I'm not sure that we'll agree on its value at a design level. However I've decided that maintaining my fork separately is doing more harm than good, and need to open the serious discussion)

Allows a user to set an environment variable `RMW_IMPLEMENTATION_WRAPPER`, in addition to `RMW_IMPLEMENTATION`.

This variable will pre-load an additional shared library that acts as the first pass of the `rwm` implementation. However, instead of just being called, this library also receives function pointers for passthrough to the "real" `RMW_IMPLEMENTATION`

This feature allows plugging in some "middleware-middleware" to do arbitrary processing on calls going through RMW at that layer, before going on to the implementation, without depending on changing `rcl`, or locking in to any of the implementations (Fast-DDS, Cyclone, Zenoh, etc))

EDIT: removed most of the disussion on Topic Statistics

> Use case note: I have been using this feature stably on a fork for about 2 years in production to enable [`rmw_stats_shim`](https://github.com/ros-tooling/graph-monitor/tree/main/rmw_stats_shim) which intercepts calls to `rmw_publish` and `rmw_receive` to compute and publish topic statistics on all publishers and subscriptions throughout an entire application. That library is not part of this review and is only mentioned to illustrate the type of thing this PR enables.

### Is this user-facing behavior change?

Optionally.

### Did you use Generative AI?

No

### DIFF
--- a/rmw_implementation/src/functions.cpp
+++ b/rmw_implementation/src/functions.cpp
@@ -82,6 +82,19 @@ load_library()
   // 3. If that fails, try loading all other implementations available in turn
   //    until one succeeds or we run out of options.
 
+  std::string wrapper_var;
+  try {
+    wrapper_var = rcpputils::get_env_var("RMW_IMPLEMENTATION_WRAPPER");
+  } catch (const std::exception & e) {
+    RMW_SET_ERROR_MSG_WITH_FORMAT_STRING(
+      "failed to fetch RMW_IMPLEMENTATION_WRAPPER "
+      "from environment due to %s", e.what());
+    return nullptr;
+  }
+  if (!wrapper_var.empty()) {
+    return attempt_to_load_one_rmw(wrapper_var);
+  }
+
   std::string env_var;
   try {
     env_var = rcpputils::get_env_var("RMW_IMPLEMENTATION");
--- a/rmw_implementation/src/functions.cpp
+++ b/rmw_implementation/src/functions.cpp
@@ -82,19 +82,6 @@ load_library()
   // 3. If that fails, try loading all other implementations available in turn
   //    until one succeeds or we run out of options.
 
-  std::string wrapper_var;
-  try {
-    wrapper_var = rcpputils::get_env_var("RMW_IMPLEMENTATION_WRAPPER");
-  } catch (const std::exception & e) {
-    RMW_SET_ERROR_MSG_WITH_FORMAT_STRING(
-      "failed to fetch RMW_IMPLEMENTATION_WRAPPER "
-      "from environment due to %s", e.what());
-    return nullptr;
-  }
-  if (!wrapper_var.empty()) {
-    return attempt_to_load_one_rmw(wrapper_var);
-  }
-
   std::string env_var;
   try {
     env_var = rcpputils::get_env_var("RMW_IMPLEMENTATION");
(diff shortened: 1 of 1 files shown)


ros-discourse:
This pull request has been mentioned on **Open Robotics Discourse**. There might be relevant details there:

https://discourse.openrobotics.org/t/ros-pmc-minutes-for-october-14-2025/50533/1

peci1:
I really like the idea of getting pub/sub stats for free. However, as you say, logically, this thing seems to stand aside a bit... Have you tested the performance impacts of using the wrapped symbols? I guess that with no wrapper, the impact is zero?

fujitatomoya (commented):
just out of curiosity, are you trying to replace `TopicStatisticsOptions` in rclcpp with this feature?

I am not sure if the topic is the user frontend to get the statistics... this also goes to `TopicStatisticsOptions` in rclcpp. (which btw, I do not really like the design of `TopicStatisticsOptions` in rclcpp either because of the following reasons.)

- in the real application, it queries the statistics to adjust the parameter in real time. searching the appropriate topic or the concerned data from the giant topic would not be ideal to react in real time application. i think it makes more sense to provide the statistics API on the endpoint (means that user application can call the statistics API on the endpoint such as publisher and subscription), so that the application can get the statistics on the endpoint and adjust the parameter in real time for process local scope.
- for the ROS 2 system statistics or observability, how about creating the service (which should be configurable) to pull the statistics for the node as it needs to do (i am suggesting the pull design based on service instead of push design based on topics)? so that collector can scrape the statistics as it needs to do based on the each endpoint requirement, without getting interrupted by unconcerned statistics? basically i think that it is not up to the application whether it needs to publish the statistics or not, it is system application to get the statistics as it needs to. (that said, the application should always expose the statistics ready.)
- or maybe, push and pull design can be configurable by system?

emersonknapp:
@peci1 I haven't been able to think of a very good way to test the overhead. Removing consideration of what the wrapper is doing with each call - because this is fully dependent on the implementation details of the loaded wrapper library - the overhead of the redirect is simply that: a redirect, one extra function (pointer) call. With no wrapper loaded, the added overhead is not "zero" - there is an extra `if()` statement added to check if the wrapper function is set. But it is effectively zero.

@fujitatomoya these are good discussion points - but I actually want to avoid having that discussion here. Mentioning it in the description was a risk.

This PR creates a generic way to intercept calls through the RMW API with a shim library.
I want to make sure that we review it in that context.

Topic Statistics via the `rmw_stats_shim` library is one way I am using this feature. But, I am _not_ proposing here or anywhere else that we consider that library for standardization or review in a ROS 2 core context. That's a conversation for later in a different place.

emersonknapp:
I have updated the PR description to reflect my wish to leave Topic Statistics out of the review discussion

ros-discourse:
This pull request has been mentioned on **Open Robotics Discourse**. There might be relevant details there:

https://discourse.openrobotics.org/t/ros-pmc-minutes-for-november-18-2025/51101/1

mjcarroll (approved):

fujitatomoya:
(just a comment, you can ignore me if others are good to go.)

i am not positive to take this PR if that is only for statistics use case because of https://github.com/ros2/rmw_implementation/pull/266#pullrequestreview-3353897632. (we have been through many experience for real-time streaming / broadcasting distributed application, calling statistics API on the endpoint is necessary with low latency for distributed system.) i would suggest to design the statistics API for ROS 2 application not only for rclcpp 1st, and tooling application can rely on the generic ROS 2 statistics APIs to do whatever needs to be done instead of opening the door like this design lets the application depends on `wrapped rmw` interfaces. (can be also maintenance issue, that says if we change the rmw signature that affects `wrapped rmw` and possibly break user space.)

so the question is that what else we actually have in these `wrap rmw` doors??? 🤔 

and if we have this door open, i believe we need to maintain the `wrap rmw` interfaces in the mainline just like https://github.com/ros2/rmw as rmw standard with documentation.

mjcarroll:
This is good feedback @fujitatomoya.  We can wait for @emersonknapp to get back to address your concerns here.

MiguelCompany:
I like this idea but:

1. I think it should be enabled with a cmake option, which would be off by default
2. We should consider what to do when building with static link (for instance, only allowing the mechanism when `BUILD_SHARED_LIBS` is `ON`)

emersonknapp:
This is all fair thoughtful feedback - I've been away on parental leave but I'm back now and would like to not let this go stale.

Yes, as discussed, my primary motivation is topic statistics. I believe it's possible to implement in core, but it would need to be at the `rcl` or `rmw` level - and I really don't want to have to do the implementation in C. But like I said, this is a separate conversation and I will open a new place to discuss it, perhaps in `ros2/design`

emersonknapp:
Okay, after much discussion we've decided to move the topic statistics functionality under `rcl`, allowing for a C++ implementation library used underneath the pure-C interface. 
Timeline now is post-Lyrical. I'll close out this PR given I don't have another use case for the shim.

Thanks everyone for the thoughts and discussion.

